PR #241, opened 2024-09-12
@W-16375669@  Automate updating of custom functions in migrated IP & DR

girjeshshakya:
### What does this PR do?
@W-16375669@
this replaces the old custom formula syntax to standard formula syntax for custom implementation.
for example : 
REPLACE(targetString, replacementString). ===> FUNCTION(“CustomStringFunctions”, “replace”, targetString, replacementString).

### DIFF
--- a/src/migration/dataraptor.ts
+++ b/src/migration/dataraptor.ts
@@ -6,7 +6,7 @@ import { DebugTimer, QueryTools } from '../utils';
 import { NetUtils } from '../utils/net';
 import { BaseMigrationTool } from './base';
 import { MigrationResult, MigrationTool, ObjectMapping, TransformData, UploadRecordResult } from './interfaces';
-import getReplacedformulaString, { getAllFunctionMetadata } from '../utils/formula/FormulaUtil';
+import { getAllFunctionMetadata, getReplacedString } from '../utils/formula/FormulaUtil';
 
 export class DataRaptorMigrationTool extends BaseMigrationTool implements MigrationTool {
   static readonly DRBUNDLE_NAME = 'DRBundle__c';
@@ -54,7 +54,6 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
     const dataRaptors = await this.getAllDataRaptors();
     const dataRaptorItemsData = await this.getAllItems();
     const functionDefinitionMetadata = await getAllFunctionMetadata(this.namespace, this.connection);
-
     // Start transforming each dataRaptor
     DebugTimer.getInstance().lap('Transform Data Raptor');
     let done = 0;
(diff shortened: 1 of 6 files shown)


AbhinavKumar-sf (changes requested):
Please write Unit test cases

src/migration/dataraptor.ts
AbhinavKumar-sf, line 1: Can we fix this large diff

src/migration/dataraptor.ts
```diff
@@ -6,258 +6,288 @@ import { DebugTimer, QueryTools } from '../utils';
...
+    const functionDefinitionMetadata = await getAllFunctionMetadata(this.namespace, this.connection);
+    // Start transforming each dataRaptor
+    DebugTimer.getInstance().lap('Transform Data Raptor');
+    let done = 0;
+    const total = dataRaptors.length;
+
+    if (functionDefinitionMetadata.length > 0 && dataRaptorItemsData.length > 0) {
+      // do the formula updation in the DR items
+      for (let drItem of dataRaptorItemsData) {
+        if (drItem[this.namespacePrefix + 'Formula__c'] != null) {
```
AbhinavKumar-sf: Let's catch the error, error in updating one drItem shouldn't impact other 
Also put logging for each item processing

girjeshshakya:
> Please write Unit test cases

@AbhinavKumar-sf added the unit tests

src/migration/omniscript.ts
```diff
@@ -187,6 +189,20 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
 
       // Get All elements for each OmniScript__c record(i.e IP/OS)
       const elements = await this.getAllElementsForOmniScript(recordId);
+      if (omniscript[`${this.namespacePrefix}IsProcedure__c`] === true) {
+        // do the formula replacement from custom to standard notation
+        if (functionDefinitionMetadata.length > 0 && elements.length > 0) {
+          for (let ipElement of elements) {
+            if (ipElement[`${this.namespacePrefix}PropertySet__c`] != null) {
```
AbhinavKumar-sf, line 196: Shouldn't we read specific key's value of  PropertySet__c and update that rather than searching and replacing in entire PropertySet config, also explore if you can move this inside **mapElementData** method.

please add logging and error handling

ankitakumar04 (approved):

girjeshshakya:
> Please write Unit test cases

Added

src/migration/omniscript.ts
girjeshshakya: added error handling and logging.
and for inside IP it is not possible to know where the formula is being configured, so this is the only way for parse and replace the formula syntax.

AbhinavKumar-sf (approved):

PR #244, opened 2024-09-13
@w-16375935@: LWC parser automated changes, fix build errors

ghost:
### What does this PR do?
LWC parser automated changes, fix build errors

Automate LWC parser.


src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+import { BaseRelatedObjectMigration } from './BaseRealtedObjectMigration';
+
+const LWC_DIR_PATH = '/force-app/main/default/lwc';
+const LWCTYPE = 'LightningComponentBundle';
+const XML_TAG_TO_REMOVE = 'runtimeNamespace';
+const NAMESPACE = 'vlocity_ins';
+
+export class LwcMigration extends BaseRelatedObjectMigration {
+  public identifyObjects(migrationResults: MigrationResult[]): Promise<JSON[]> {
+    throw new Error('Method not implemented.');
```
AbhinavKumar-sf, line 23: Can you implement this method also to identify the changes in the files

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+  public migrateRelatedObjects(migrationResults: MigrationResult[], migrationCandidates: JSON[]): void {
+    this.migrate();
+  }
+  // eslint-disable-next-line @typescript-eslint/member-ordering
+
+  public migrate(): void {
+    const pwd = shell.pwd();
```
AbhinavKumar-sf, line 31: We should return the LWC name,  files names along with token ( location in the file) which ever is getting modified

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+    sfProject.deploy(LWCTYPE, targetOrg.getUsername());
+    shell.cd(pwd);
+  }
+
+  public processLwcFiles(dir: string): File[] {
+    dir += LWC_DIR_PATH;
+    let files: File[] = [];
+    // files = fileutil.readFilesSync(dir);
+    files = fileutil.readAllFiles(dir);
+    // TODO: Add logging
```
AbhinavKumar-sf: Let's log each LWC  which is getting processed,  Also log  each file getting processed and each modification we are doing

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+    for (const file of files) {
+      if (file.ext === '.js') {
```
AbhinavKumar-sf: Put this inside a method  and try catch , as error in modifying one file shouldn't block other file updation

src/utils/file/fileutil.ts
```diff
@@ -15,6 +15,32 @@ export class fileutil {
     });
     return files;
   }
+
+  public static readAllFiles(dirPath: string, fileList: File[] = []): File[] {
+    // Read the directory contents
+    const files = fs.readdirSync(dirPath);
```
AbhinavKumar-sf, line 21: What if we face error in reading directory/file,  should we handle that case?

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.processJavascriptFile(file);
+      } else if (file.ext === '.html') {
+        this.processHtmlFile(file);
+      } else if (file.ext === '.xml') {
+        this.processXMLFile(file);
+      }
```
AbhinavKumar-sf: Let's also log the file name if it is not falling under these 3 categories and we are not modifying it

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+import { sfProject } from '../../utils/sfcli/project/sfProject';
+import { JavaScriptParser } from '../../utils/lwcparser/jsParser/JavaScriptParser';
+import { HTMLParser } from '../../utils/lwcparser/htmlParser/HTMLParser';
+import { XmlParser } from '../../utils/lwcparser/xmlParser/XmlParser';
+import { BaseRelatedObjectMigration } from './BaseRealtedObjectMigration';
+
+const LWC_DIR_PATH = '/force-app/main/default/lwc';
+const LWCTYPE = 'LightningComponentBundle';
+const XML_TAG_TO_REMOVE = 'runtimeNamespace';
+const NAMESPACE = 'vlocity_ins';
```
AbhinavKumar-sf: Why this const, shouldn't this be read from property

src/utils/lwcparser/jsParser/JavaScriptParser.ts
```diff
@@ -1,49 +1,47 @@
...
-    const astString = this.getAST(); // Get the AST as a string
-    if (astString) {
-      console.log(astString); // Output the AST
-    }
+    // Generate the updated code from the modified AST
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call
+    const output = generate(ast, {}, code);
+
+    // Write the modified code back to the file
+    fs.writeFileSync(filePath, output.code, 'utf-8');
```
AbhinavKumar-sf: Let parser only implement the updation logic and let it return updated content, File write should be handled be  caller

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+      } else if (file.ext === '.html') {
+        this.processHtmlFile(file);
+      } else if (file.ext === '.xml') {
+        this.processXMLFile(file);
+      }
+    }
+    return files;
+  }
+
+  processJavascriptFile(file: File): void {
```
AbhinavKumar-sf, line 69: shouldn't this return updatedContent

src/migration/related/LwcMigration.ts
```diff
@@ -0,0 +1,79 @@
...
+    const jsParser = new JavaScriptParser();
+    const filePath = file.location;
+    const namespace = NAMESPACE;
+    jsParser.replaceImportSource(filePath, namespace);
+  }
+
+  processHtmlFile(file: File): void {
+    const filePath: string = file.location;
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
AbhinavKumar-sf: Do we need these exceptions?

src/utils/lwcparser/jsParser/JavaScriptParser.ts
AbhinavKumar-sf, line 1: Shall we not move these test inputs inside test folder instead

src/migration/related/LwcMigration.ts
ghost: This method I'm implementing in current work item, so this can be taken care.

PR #270, opened 2025-04-30
@W-18307604 feat: identify data model and package version used in the org

sf-pranav-varshney:


sf-aastha-paruthi (approved):
Please add details of changes.

sf-aastha-paruthi:
Please update the subject of PR. It needs to be in specific format for build to trigger.

sf-aastha-paruthi (changes requested):
Please check if commits need to be verified.

sf-aastha-paruthi (approved):

src/utils/orgUtils/index.ts
```diff
@@ -0,0 +1,312 @@
+/* eslint-disable */
```
mdmehran-qureshi, line 1: Can we use this method instead:

export async function getNamespaceFromOrg(conn: any): Promise<string | undefined> {
  try {
    const queryResult = await conn.query("SELECT NamespacePrefix FROM ApexClass WHERE Name = 'DRGlobal'");
    if (queryResult.records && queryResult.records.length > 0) {
      return (queryResult.records[0] as { NamespacePrefix: string }).NamespacePrefix;
    }
  } catch (error) {
    Logger.warn('Could not determine namespace from org');
    Logger.logVerbose(`Namespace query error: ${error}`);
  }
  return undefined;
}

mdmehran-qureshi (changes requested):

src/utils/orgUtils/index.ts
sf-pranav-varshney: Thanks @mdmehran-qureshi Really appreciate the solution you have suggested. 

but it is not 100% full proof solution, Apex classes can have the same name if they belong to different managed packages then what would be out source of truth.

Instead we directly queried in Publisher table.

sf-pranav-varshney: You're absolutely right  and great observation.

While using the namespace prefix helps distinguish between Apex classes with the same name across managed packages, it’s not a 100% foolproof solution for uniquely identifying the source or publisher of a class, especially when dealing with cross-package conflicts.

Querying the Publisher metadata gives you a more reliable and authoritative source of truth about "Who owns the package", 'Which version of the package is installed" and "What namespace is used".

src/utils/orgUtils/index.ts
```diff
@@ -0,0 +1,312 @@
...
+    const allInstalledPackages = (await QueryTools.queryAll(
+      connection,
+      '',
+      this.objectName,
+      this.fields
+    )) as unknown as InstalledPackage[];
+
+    const packageDetails: PackageDetail[] = allInstalledPackages
+      // Filter packages to only include those with a namespace in the predefined list
+      .filter((pkg) => this.namespaces.has(pkg.NamespacePrefix))
```
mdmehran-qureshi, line 283: .filter((pkg) => namespace === pkg.NamespacePrefix) ?

mdmehran-qureshi (approved):
lgtm

PR #280, opened 2025-06-04
Add Support for Only and RelatedObjects Flags in Assess Mode

rajender-kumar-salesforce:


src/utils/prototypes.ts
rajender-kumar-salesforce, line 1: This file and its imported were removed because of Test case failures. Added it as util instead of prototype.

src/commands/omnistudio/migration/assess.ts
```diff
@@ -39,6 +37,10 @@ export default class Assess extends OmniStudioBaseCommand {
       description: messages.getMessage('allVersionsDescription'),
       required: false,
     }),
+    relatedobjects: flags.string({
+      char: 'r',
+      description: messages.getMessage('apexLwc'),
```
sf-pranav-varshney, line 41: Do you think, we need to change the message as the existing apexlwc message is specific for migration. Please check once.

src/commands/omnistudio/migration/assess.ts
```diff
@@ -61,41 +65,122 @@ export default class Assess extends OmniStudioBaseCommand {
...
+      // If no specific component is specified, assess all components
+      await this.assessDataRaptors(assesmentInfo, namespace, conn);
+      await this.assessFlexCards(assesmentInfo, namespace, conn, allVersions);
+      await this.assessOmniScripts(assesmentInfo, namespace, conn, allVersions, OmniScriptExportType.All);
+      return;
+    }
+
+    switch (assessOnly) {
+      case 'dr':
+        await this.assessDataRaptors(assesmentInfo, namespace, conn);
```
sf-pranav-varshney, line 129: Since the function doesn't do anything explicitly and just assign the result, we can avoid the extra function call. I suggest simplifying it like this:

`assessmentInfo.dataRaptorAssessmentInfos = await new DataRaptorMigrationTool(namespace, conn, this.logger, messages, this.ux).assess();`

What do you think?

src/migration/omniscript.ts
```diff
@@ -103,7 +104,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
 
     let success: boolean = await NetUtils.delete(this.connection, ids);
     if (!success) {
-      throw new Error(this.messages.getMessage('couldNotTruncateOmnniProcess').formatUnicorn(objectName));
+      throw new Error(formatUnicorn(this.messages.getMessage('couldNotTruncateOmnniProcess'), objectName));
```
sf-pranav-varshney, line 107: Why it is required?

src/migration/related/OmnistudioRelatedObjectMigrationFacade.ts
```diff
@@ -108,7 +107,48 @@ export default class OmnistudioRelatedObjectMigrationFacade {
...
+    // Return results needed for --json flag
+    return { apexAssessmentInfos, lwcAssessmentInfos };
+  }
+
+  public assessAll(relatedObjects: string[]): RelatedObjectAssesmentInfo {
+    // Start the debug timer
+    DebugTimer.getInstance().start();
+
+    // Declare an array of MigrationTool
+    const projectDirectory: string = OmnistudioRelatedObjectMigrationFacade.intializeProject();
```
sf-pranav-varshney: I think, projectDirectory should be initialize inside construct and class variable. I know it is same as MIgration but i don't think it is right design. Lets discuss and fix it.

src/migration/omniscript.ts
rajender-kumar-salesforce: @sf-pranav-varshney There were test case failures in local after adding formatUnicorn in string prototype. I have moved this method into util. Also respective prototype file `src/utils/prototypes.ts`. We also get unnecessary warnings.

Screenshot 2025-06-10 at 12 25 32 PM

src/commands/omnistudio/migration/assess.ts
rajender-kumar-salesforce: Nice catch. fixed message description.

sf-pranav-varshney (approved):
Look good to me. Please test migration as well.

sf-pranav-varshney:
@rajender-kumar-salesforce Please add the relatedObject in ReadMe as well.

snehaljha-sf (approved):

rajender-kumar-salesforce:
> Look good to me. Please test migration as well.

Verified for migration. Did minor fix wrt variable initialization.

PR #298, opened 2025-06-24
Multiple Omnistudio packages in single org & lint resolve

mdmehran-qureshi:


sf-pranav-varshney (approved):

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -101,7 +101,7 @@ export default class Migrate extends OmniStudioBaseCommand {
       await OrgPreferences.enableOmniPreferences(conn);
     } catch (error) {
       const errMsg = error instanceof Error ? error.message : String(error);
-      this.ux.log(`Could not enable Omni preferences: ${errMsg}`);
+      Logger.log(`Could not enable Omni preferences: ${errMsg}`);
```
sf-pranav-varshney, line 104: logger.error?

mdmehran-qureshi: I am using Logger class directly. From each static method of Logger class, we are writing logs in log file.

PR #304, opened 2025-06-26
@W-18814957: Error handling and issue fixes

snehaljha-sf:


src/commands/omnistudio/migration/assess.ts
```diff
@@ -50,13 +50,25 @@ export default class Assess extends OmniStudioBaseCommand {
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   public async run(): Promise<any> {
+    Logger.initialiseLogger(this.ux, this.logger, 'assess', this.flags.verbose);
+    try {
```
sf-pranav-varshney, line 54: As discussed, the following changes are needed as part of Label PR.

1) Please create the logger as a singleton class to ensure it's used consistently across the code.
2) We also need to improve the log messages. Instead of passing them directly as function arguments, let's define them in a constants class and use those constants throughout the code.
3) As dicussed, Please change the Logger error which accept message and trace together.

sf-pranav-varshney (approved):

src/commands/omnistudio/migration/assess.ts
```diff
@@ -128,7 +140,7 @@ export default class Assess extends OmniStudioBaseCommand {
       Logger.log((error as Error).message);
       Logger.log((error as Error).stack);
     }
-    await AssessmentReporter.generate(assesmentInfo, conn.instanceUrl, orgs, assessOnly, objectsToProcess);
+    await AssessmentReporter.generate(assesmentInfo, conn.instanceUrl, orgs, assessOnly, objectsToProcess, messages);
```
sf-kishore-kurri, line 143: message arg passing will change post the singleton implementation right?

snehaljha-sf: yes, it has been fixed in the next PR

sf-kishore-kurri (commented):
added comments but these will be taken care in next PR

sf-kishore-kurri (approved):

PR #312, opened 2025-07-01
Test fix

snehaljha-sf:


src/commands/omnistudio/migration/assess.ts
```diff
@@ -221,6 +137,7 @@ export default class Assess extends OmniStudioBaseCommand {
     let objectsToProcess: string[];
     // Assess related objects if specified
     if (relatedObjects) {
+      const projectPath = await this.getProjectPath();
```
mdmehran-qureshi: Call it at line number 117 only.
Reason: We don't wan't manual intervention from user after other object assessment. In this case it will first assess DR, OS/IP and FC then ask user for project path. It is not a good scenario.

mdmehran-qureshi (approved):

src/commands/omnistudio/migration/assess.ts
snehaljha-sf: got it

sf-pranav-varshney (approved):

PR #319, opened 2025-07-03
feat: @W-18541630: Apex remote and Custom LWC dependencies support for Flexcard

sf-kishore-kurri:

### DIFF
--- a/src/migration/flexcard.ts
+++ b/src/migration/flexcard.ts
@@ -133,6 +133,7 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
           dependenciesFC: [],
           dependenciesOS: [],
           dependenciesLWC: [],
+          dependenciesApexRemoteAction: [],
           infos: [],
           warnings: [this.messages.getMessage('unexpectedError')],
         });
@@ -156,6 +157,7 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
       dependenciesOS: [],
       dependenciesFC: [],
       dependenciesLWC: [],
+      dependenciesApexRemoteAction: [],
       infos: [],
       warnings: [],
     };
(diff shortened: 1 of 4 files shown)


src/migration/flexcard.ts
```diff
@@ -353,25 +359,14 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
-          const lwcName = lwcData.NamespacePrefix
-            ? `${lwcData.NamespacePrefix}.${lwcData.DeveloperName}`
-            : lwcData.DeveloperName;
-
-          // Avoid duplicates
-          if (!flexCardAssessmentInfo.dependenciesLWC.includes(lwcName)) {
-            flexCardAssessmentInfo.dependenciesLWC.push(lwcName);
-          }
+      // Check customlwcname property first
+      if (component.property.customlwcname) {
```
sf-pranav-varshney, line 368: Is it possible that component.property can be undefined or null? If yes then it better to add check.

src/migration/flexcard.ts
```diff
@@ -353,25 +359,14 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+        const customLwcName = component.property.customlwcname;
+        Logger.log(`Custom LWC name: ${customLwcName}`);
```
sf-pranav-varshney: use verbose instead log.

src/migration/flexcard.ts
```diff
@@ -237,6 +239,10 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
           );
         }
       }
+    } else if (dataSource.type === Constants.ApexRemoteComponentName) {
+      const remoteClass = dataSource.value?.remoteClass;
+      const remoteMethod = dataSource.value?.remoteMethod;
+      flexCardAssessmentInfo.dependenciesApexRemoteAction.push(`${remoteClass}.${remoteMethod}`);
```
sf-pranav-varshney: Do we need to handle duplicate here as well similar to LWC?

sf-pranav-varshney (approved):
The changes look good to me. Please address the review comments if you think they are necessary.

src/migration/flexcard.ts
sf-kishore-kurri: @sf-pranav-varshney  we are checking in if clause
`if (component.element === 'customLwc' && component.property) {`

src/migration/flexcard.ts
sf-kishore-kurri: Addressed

src/migration/flexcard.ts
sf-kishore-kurri: Addressed

sf-aastha-paruthi (approved):

PR #324, opened 2025-07-04
Fix: Warning Message and Dashboard titles

rajender-kumar-salesforce:


messages/migrate.json
```diff
@@ -94,5 +94,6 @@
   "errorParsingData": "Error while parsing data for template",
   "errorGeneratingHTML": "Error while generating HTML from template for {} with properties {}",
   "errorEvaluatingExpression": "Error evaluating expression: {0}, {1}",
-  "manuallySwitchDesignerToStandardDataModel": "To complete the setup, please manually run the required Apex code to configure OmniStudio Designers to use the Standard Data Model <br> https://help.salesforce.com/s/articleView?id=xcloud.os_migrate_change_the_sobject_data_model_after_migration.htm&type=5"
+  "manuallySwitchDesignerToStandardDataModel": "To complete the setup, please manually run the required Apex code to configure OmniStudio Designers to use the Standard Data Model <br> https://help.salesforce.com/s/articleView?id=xcloud.os_migrate_change_the_sobject_data_model_after_migration.htm&type=5",
+  "nameConversionMessage": "WARNING: {0} name has been modified to fit naming rules: {1}"
```
sf-kishore-kurri: {0} is not working so we should use %s for replacement

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -383,7 +383,12 @@ export default class Migrate extends OmniStudioBaseCommand {
         obj.status = !recordResults || recordResults.hasErrors ? 'Error' : 'Complete';
         obj.errors = errors;
         obj.migratedId = recordResults.id;
-        obj.warnings = recordResults.warnings;
+        obj.warnings = recordResults.warnings || [];
+        if (recordResults.newName !== obj.name) {
+          const entityGroup = migrationResults.name;
```
sf-kishore-kurri: I forgot the usecase, what is this change for @rajender-kumar-salesforce

rajender-kumar-salesforce: Warning message once OS, FC, DM, IP naming has been changed after migration.

messages/migrate.json
rajender-kumar-salesforce: Done

sf-kishore-kurri (approved):
lgtm

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -383,7 +383,12 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+          const entity = entityGroup === 'Data Mapper' ? entityGroup : entityGroup.substring(0, entityGroup.length - 1);
```
snehaljha-sf: 1. Are we completely sure about getting name parameter for all the components?
2. I see the warning has been removed from omniscript but wouldn't we get same warnings appended twine for other components like flexcard

rajender-kumar-salesforce: 1. Yes we get name for every component.
2. For IP, OS warning message was appearing even their new name and old name are same. Moved this comparison and warning message generation from OS/IP.

sf-kishore-kurri: Did you check FC and DR usecase also?

PR #328, opened 2025-07-10
@W-18480490 - Call to Action link Validation

shaurabh-tiwari-git:

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "jsdom": "^25.0.0",
     "lodash.chunk": "^4.2.0",
     "open": "^8.4.2",
+    "puppeteer": "^24.12.1",
     "shelljs": "^0.8.5",
     "tslib": "^2",
     "xmldom": "^0.6.0"
--- a/test/utils/constants/documentRegistry.test.ts
+++ b/test/utils/constants/documentRegistry.test.ts
@@ -1,0 +1,85 @@
+import puppeteer from 'puppeteer';
+import { expect } from '@salesforce/command/lib/test';
+import { documentRegistry } from '../../../src/utils/constants/documentRegistry';
+import { Logger } from '../../../src/utils/logger';
+
+describe('DocumentRegistry', () => {
+  describe('URL Validation', () => {
+    // Helper function to make HTTP request and check if URL is accessible
+    async function checkSalesforceUrlWithPuppeteer(url: string): Promise<boolean> {
+      const browser = await puppeteer.launch({ headless: true }); // use true instead of 'new'
+      const page = await browser.newPage();
+      try {
+        await page.goto(url, { waitUntil: 'networkidle2', timeout: 15000 });
+        const content = await page.content();
+        const notFoundText = "couldn't find that page.";
+        const isValid = !content.includes(notFoundText);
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -18,6 +18,7 @@
     "jsdom": "^25.0.0",
     "lodash.chunk": "^4.2.0",
     "open": "^8.4.2",
+    "puppeteer": "^24.12.1",
```
sf-aastha-paruthi, line 21: Does the same package need to be added in yarn

sf-aastha-paruthi:
LGTM

test/utils/constants/documentRegistry.test.ts
```diff
@@ -0,0 +1,84 @@
...
+import { expect } from '@salesforce/command/lib/test';
+import { documentRegistry } from '../../../src/utils/constants/documentRegistry';
+import { Logger } from '../../../src/utils/logger';
+
+describe('DocumentRegistry', () => {
+  describe('URL Validation', () => {
+    // Helper function to make HTTP request and check if URL is accessible
+    async function checkSalesforceUrlWithPuppeteer(url: string): Promise<boolean> {
+      const browser = await puppeteer.launch({ headless: true }); // use true instead of 'new'
+      const page = await browser.newPage();
```
sf-kishore-kurri, line 49: Do we need puppeteer here, can't we simply use requests http library to hint the link and check for 200 response code?

shaurabh-tiwari-git: No, the links are not returning the response as 200, they are returning response 301, for valid as well as invalid URLs. And on the redirected URL the error is not coming directly, javascript rendering is first happening on the UI and then the error is showing up, which can not we captured without a browser environment.

package.json
sf-kishore-kurri: Did you validate the use case with plugin as well?

test/utils/constants/documentRegistry.test.ts
sf-kishore-kurri: @shaurabh-tiwari-git as discussed add the title check assertion also

yarn.lock
```diff
@@ -11,13 +11,13 @@
 
 "@apexdevtools/apex-parser@^4.1.0":
   version "4.1.0"
-  resolved "https://registry.yarnpkg.com/@apexdevtools/apex-parser/-/apex-parser-4.1.0.tgz#f8e801a8e4a65ff0e1fb70629642cda76e60fed3"
+  resolved "https://registry.npmjs.org/@apexdevtools/apex-parser/-/apex-parser-4.1.0.tgz"
```
sf-kishore-kurri: Do we need to check-in this file or it will be auto built based on package.json changes?  Did you check with @snehaljha-sf about it?

sf-kishore-kurri (approved):
lgtm, but lets check whether we should check-in yarn.lock or not

yarn.lock
snehaljha-sf: This one is alright @sf-kishore-kurri, it'll ensure we have the exact dependency that have been added during development and validated

snehaljha-sf (approved):

PR #343, opened 2025-07-23
W-19107238: Fix Global Auto Number UI Error Labels

rajender-kumar-salesforce:


src/commands/omnistudio/migration/migrate.ts
```diff
@@ -165,6 +165,9 @@ export default class Migrate extends OmniStudioBaseCommand {
     const debugTimer = DebugTimer.getInstance();
     // We need to truncate the standard objects first
     let objectMigrationResults = await this.truncateObjects(migrationObjects, debugTimer);
+    objectMigrationResults = objectMigrationResults.filter(
+      (result) => result.name !== Constants.GlobalAutoNumberComponentName
```
sf-kishore-kurri, line 169: Is this logic to handle the truncate scenario? @rajender-kumar-salesforce

src/utils/OmniGlobalAutoNumberPrefManager.ts
```diff
@@ -35,7 +35,7 @@ export class OmniGlobalAutoNumberPrefManager {
       return result?.enableOmniGlobalAutoNumberPref === 'true' || false;
     } catch (error) {
       const errMsg = error instanceof Error ? error.message : String(error);
-      Logger.error(`Error checking OmniGlobalAutoNumberPref status: ${errMsg}`);
+      Logger.error(`We couldn’t check whether the Global Auto Number setting is enabled: ${errMsg}. Try again later.`);
```
sf-kishore-kurri: we should read this string from messages here

src/migration/globalautonumber.ts
```diff
@@ -45,12 +46,11 @@ export class GlobalAutoNumberMigrationTool extends BaseMigrationTool implements
   }
 
   public async truncate(): Promise<void> {
-    try {
-      // Perform pre-migration checks before truncation
-      await this.performPreMigrationChecks();
+    // Perform pre-migration checks before truncation
+    await this.performPreMigrationChecks();
+    this.globalAutoNumberSettings = await this.getAllGlobalAutoNumberSettings();
```
sf-kishore-kurri, line 51: what settings we are checking here?

src/commands/omnistudio/migration/migrate.ts
rajender-kumar-salesforce: Correct. Even if truncation for global auto number fails we don't want block user from migrating other objects.

src/migration/globalautonumber.ts
rajender-kumar-salesforce: Before truncation checking if package has data or not. If package doesn't have data then don't truncate objects from core.

src/utils/logger.ts
```diff
@@ -65,9 +65,7 @@ export class Logger {
       } else {
         Logger.sfUX.error(`\x1b[31m${message}\x1b[0m`);
         if (error) {
-          Logger.sfUX.error(`\x1b[31m${message}\n${error.message}\n${error.stack}\x1b[0m`);
-        } else {
-          Logger.sfUX.error(`\x1b[31m${message}\x1b[0m`);
+          Logger.sfUX.error(`\x1b[31m${error.message}\n${error.stack}\x1b[0m`);
```
rajender-kumar-salesforce, line 68: Fix for duplicate error messages shown on terminal @sf-kishore-kurri

src/utils/OmniGlobalAutoNumberPrefManager.ts
rajender-kumar-salesforce: Done

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: @rajender-kumar-salesforce But Do we need to truncate the globalAutoNumber records is my question?

rajender-kumar-salesforce: If we don't truncate:
Scenario: user runs migration for global auto number and it migrated objects but perf enabelemt fails. Now user goes to package and updates any global auto number field. 
Now if user runs migration for global auto number then certain records will be duplicates. We need to update core object field for these records in core as well right ? We have to either update, delete those objects in core and recerate it. But here also we are deleting objects from core.

sf-kishore-kurri: @rajender-kumar-salesforce When we doing migration ideally we shouldn't create any new data on old system post migration

sf-kishore-kurri (approved):
Let's discuss Truncate change, but remaining changes looks good to me

PR #344, opened 2025-07-25
fix: more mvp issues

snehaljha-sf:

### DIFF
--- a/src/commands/omnistudio/migration/migrate.ts
+++ b/src/commands/omnistudio/migration/migrate.ts
@@ -14,7 +14,7 @@ import { ExecuteAnonymousResult } from 'jsforce';
 import OmniStudioBaseCommand from '../../basecommand';
 import { DataRaptorMigrationTool } from '../../../migration/dataraptor';
 import { DebugTimer, MigratedObject, MigratedRecordInfo } from '../../../utils';
-import { MigrationResult, MigrationTool } from '../../../migration/interfaces';
+import { InvalidEntityTypeError, MigrationResult, MigrationTool } from '../../../migration/interfaces';
 import { ResultsBuilder } from '../../../utils/resultsbuilder';
 import { CardMigrationTool } from '../../../migration/flexcard';
 import { OmniScriptExportType, OmniScriptMigrationTool } from '../../../migration/omniscript';
@@ -117,6 +117,13 @@ export default class Migrate extends OmniStudioBaseCommand {
       Logger.log(`Could not enable Omni preferences: ${errMsg}`);
     }
 
+    // check for confirmation over assessed action items
+    const migrationConsent = await this.getMigrationConsent();
+    if (!migrationConsent) {
+      Logger.log(messages.getMessage('migrationConsentNotGiven'));
+      return;
+    }
+
     const namespace = orgs.packageDetails.namespace;
     // Let's time every step
     DebugTimer.getInstance().start();
@@ -178,7 +185,9 @@ export default class Migrate extends OmniStudioBaseCommand {
     );
 
     let actionItems = [];
-    actionItems = await this.setDesignersToUseStandardDataModel(namespace);
+    if (!migrateOnly) {
+      actionItems = await this.setDesignersToUseStandardDataModel(namespace);
+    }
 
     await ResultsBuilder.generateReport(
       objectMigrationResults,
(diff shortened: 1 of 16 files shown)


src/commands/omnistudio/migration/migrate.ts
```diff
@@ -194,6 +203,38 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+
+    while (consent === null) {
+      try {
+        // Race between the confirmation prompt and the timeout
+        consent = await Promise.race([Logger.confirm(messages.getMessage('migrationConsentMessage')), timeoutPromise]);
+        clearTimeout(timeoutHandle);
+      } catch (error) {
+        clearTimeout(timeoutHandle);
+        if (error instanceof Error && error.message === messages.getMessage('requestTimedOut')) {
+          Logger.log(messages.getMessage('requestTimedOut'));
```
sf-kishore-kurri: With timeout, consent is assumed as NO and do we continue with migration? @snehaljha-sf

sf-kishore-kurri (approved):
LGTM, but please merge the same changes into GA branch post testing

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -178,7 +185,9 @@ export default class Migrate extends OmniStudioBaseCommand {
     );
 
     let actionItems = [];
-    actionItems = await this.setDesignersToUseStandardDataModel(namespace);
+    if (!migrateOnly) {
```
sf-aastha-paruthi, line 190: Why has this check been introduced? If customer with only DM wants to migrate, and one with all components wants to migrate, the flow should ideally remain same.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -194,6 +203,38 @@ export default class Migrate extends OmniStudioBaseCommand {
     return { objectMigrationResults };
   }
 
+  private async getMigrationConsent(): Promise<boolean> {
+    const TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes timeout
```
sf-aastha-paruthi: Prompt util has been created for timeout related consents. That can be reused.

src/migration/omniscript.ts
```diff
@@ -523,21 +534,20 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       if (omniProcessType === 'OmniScript') {
         const type = omniscript[this.namespacePrefix + 'IsLwcEnabled__c'] ? 'LWC' : 'Angular';
         if (type === 'Angular') {
-          // Skip Angular OmniScripts and add a warning record
-          const warningMessage = this.messages.getMessage('angularOmniscriptWarningMessage');
+          // Skip Angular OmniScripts and add a warning record this
```
sf-aastha-paruthi: nit, remove this at the end

src/migration/omniscript.ts
```diff
@@ -523,21 +534,20 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+
           const skippedResponse: UploadRecordResult = {
             referenceId: recordId,
             id: '',
-            success: false,
-            hasErrors: true,
-            errors: [warningMessage],
-            warnings: [],
+            success: true,
```
sf-aastha-paruthi: Why is false being changed to true for angluar omniscripts as the migration is not being taken care?

src/commands/omnistudio/migration/migrate.ts
snehaljha-sf: It is as per the discussion

snehaljha-sf: We don't want user to switch to standard data model if partial migration has happened

src/migration/omniscript.ts
snehaljha-sf: This won't matter now but we should default to false for success so have changed it

src/commands/omnistudio/migration/migrate.ts
snehaljha-sf: moved to already present consent flow

src/commands/omnistudio/migration/migrate.ts
sf-aastha-paruthi: If user has only dm then it is not partial migration for  that user

sf-aastha-paruthi (approved):

PR #352, opened 2025-08-01
fix: flexipage issues

snehaljha-sf:


sf-kishore-kurri (approved):
lgtm

src/utils/flexipage/flexiPageTransformer.ts
```diff
@@ -142,7 +148,7 @@ function createNewPropsForOmniScript(
 
   return {
     componentName: targetComponentNameOS,
-    identifier: targetIdentifierOS,
+    identifier: `${targetIdentifierOS}${osSeq++}`,
```
sf-aastha-paruthi, line 151: There could be a possibility that if the omniscript is created directly using standard wrapper, the identifier is generated in a specific format like 16 digit hexadecimal and the same format id might be used by others. 
Appending sequence number might change the format. 
Please check it does not break any use case and the desired format.

sf-aastha-paruthi:
Please update the description for better clarity as to which issue this PR addresses or attach the work-item containing the details.

src/utils/flexipage/flexiPageTransformer.ts
snehaljha-sf: I've checked the flexipage in core which follows same convention

sf-sayeeduddin-syed (approved):

PR #353, opened 2025-08-01
feat: @W-18480458: Implementation of update references across Omnistudio components

sf-kishore-kurri:
Changes:
- Handling references updates post migration clean up changes
- Added logic to perform migrate in correct order DM -> IP -> OS and FC
- Created the namingRegistry to store old  and cleanup name for lookup
- Added the test cases


sf-aastha-paruthi:
In case of OS and FC, both can have dependencies on each other. Is the correct order change mentioned taking care of that.

sf-kishore-kurri:
> e correct order change mentioned taking care

The order of migration happens like DM, IP, OS and FC so if OS has any FC dependency it will be taken care by the namingRegistry because it has the oldName <-> cleanedName so it will update automatically.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -440,6 +494,114 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+      Logger.log(messages.getMessage('completeComponentMappingMessage', [allMappings.length]));
+    } catch (error) {
+      Logger.error(messages.getMessage('errorComponentMapping'), error);
+    }
+  }
+
+  /**
+   * Query DataMappers from the org
+   */
+  private async queryDataMappers(conn: any, namespace: string): Promise<any[]> {
```
sf-aastha-paruthi, line 507: The functions to query and fetch the datamappers and other components are already written in each class like getAllDatamappers.
Please reuse those. A bit refactoring might be needed.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -440,6 +494,114 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    const result = await conn.query(query);
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return result.records || [];
+  }
+
+  /**
+   * Query OmniScripts/Integration Procedures from the org
+   */
+  private async queryOmniScripts(conn: any, namespace: string, isProcedure: boolean): Promise<any[]> {
```
sf-aastha-paruthi, line 518: Please reuse the already existing functions.

src/migration/NameMappingRegistry.ts
```diff
@@ -0,0 +1,401 @@
...
+    this.integrationProcedureMappings.clear();
+    this.flexCardMappings.clear();
+  }
+
+  /**
+   * Pre-process all components to register their name mappings
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public preProcessComponents(
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sf-aastha-paruthi, line 270: Please try avoiding es lint checks

src/migration/flexcard.ts
```diff
@@ -773,81 +775,261 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+  /**
+   * Update dataSource (DataRaptor, Integration Procedures, Apex Remote) using registry
+   */
+  private updateDataSourceWithRegistry(dataSource: any, invalidIpNames: Map<string, string>, context: string): void {
+    const type = dataSource.type;
+
+    if (type === Constants.DataRaptorComponentName || type === 'DataRaptor') {
+      // Handle DataRaptor using registry
+      const originalBundle = dataSource.value?.bundle || '';
+      if (originalBundle && this.nameRegistry.hasDataMapperMapping(originalBundle)) {
```
sf-aastha-paruthi, line 925: In what scenarios will the nameRegistry might not have the nameMapping.

src/migration/flexcard.ts
```diff
@@ -773,81 +775,261 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+        dataSource.value.bundle = this.nameRegistry.getDataMapperCleanedName(originalBundle);
+      } else {
+        dataSource.value.bundle = this.cleanName(originalBundle);
```
sf-aastha-paruthi, line 929: Does the nameregistry need to be updated?

src/migration/NameMappingRegistry.ts
```diff
@@ -0,0 +1,401 @@
...
+  fieldPath: string; // JSON path to the field that needs updating
+  referencedOriginalName: string;
+  referencedComponentType: string;
+}
+
+/**
+ * Centralized registry for tracking name mappings and managing dependency updates
+ * during Omnistudio component migration.
+ */
+export class NameMappingRegistry {
```
sf-aastha-paruthi, line 30: What is the use case of this NameMappingRegistry.

src/migration/NameMappingRegistry.ts
```diff
@@ -0,0 +1,401 @@
...
+
+  /**
+   * Register a component name mapping before migration
+   */
+  public registerNameMapping(mapping: ComponentNameMapping): void {
+    const key = `${mapping.componentType}:${mapping.originalName}`;
+    this.nameMappings.set(key, mapping);
+
+    // Store in type-specific maps for quick lookup
+    switch (mapping.componentType) {
```
sf-aastha-paruthi, line 62: The same can be derived at runtime. Why do we need to store the same.

sf-aastha-paruthi:
> > e correct order change mentioned taking care
> 
> The order of migration happens like DM, IP, OS and FC so if OS has any FC dependency it will be taken care by the namingRegistry because it has the oldName <-> cleanedName so it will update automatically.

This will not ensure the correct reference update. The unmigrated or Error items will not be handled correctly.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -128,58 +129,44 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+
     // Migrate individual objects
     const debugTimer = DebugTimer.getInstance();
-    // We need to truncate the standard objects first
-    let objectMigrationResults = await this.truncateObjects(migrationObjects, debugTimer);
-    objectMigrationResults = objectMigrationResults.filter(
-      (result) => result.name !== Constants.GlobalAutoNumberComponentName
-    );
+    // We need to truncate the standard objects first (in reverse order for cleanup)
+    let objectMigrationResults = await this.truncateObjects([...migrationObjects].reverse(), debugTimer);
```
sf-aastha-paruthi, line 165: Please update the name. This does not seem to be migration results.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -374,29 +408,41 @@ export default class Migrate extends OmniStudioBaseCommand {
     return objectMigrationResults;
   }
 
-  private getMigrationObjects(
+  /**
+   * Get migration objects in the correct dependency order:
+   * 1. DataMappers (lowest dependencies)
+   * 2. OmniScripts/Integration Procedures
```
sf-aastha-paruthi: Update the comment to reflect the correct order.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -420,18 +467,25 @@ export default class Migrate extends OmniStudioBaseCommand {
...
         case Constants.DataMapper:
-          migrationObjects.push(new DataRaptorMigrationTool(namespace, conn, this.logger, messages, this.ux));
+          migrationObjects.push(
+            new DataRaptorMigrationTool(namespace, conn, this.logger, messages, this.ux, nameRegistry)
+          );
           break;
         case Constants.GlobalAutoNumber:
-          migrationObjects.push(new GlobalAutoNumberMigrationTool(namespace, conn, this.logger, messages, this.ux));
+          migrationObjects.push(
+            new GlobalAutoNumberMigrationTool(namespace, conn, this.logger, messages, this.ux, nameRegistry)
```
sf-aastha-paruthi: If this is independent item, do we still have dependency on nameRegistry.

sf-kishore-kurri:
> e correct reference update. T

The failure/ error ones should be remigrated then name registry will take care of it of update references

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: For now, I'm passing the namingRegistry to GlobalAutoNumber migration process as well so that we can use it in future for any reference updates

src/migration/NameMappingRegistry.ts
sf-kishore-kurri: NamingRegistry is in memory storage which is prepared before migration for all component types to maintain key, value between oldName -> cleanName which can be used for update references across different migration component types

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: The getAllDatamappers created as private one only accessible within each migration component type and i'm preparing registry before migration so created these query utility APIs for accessing the pre-migrated data

src/migration/flexcard.ts
sf-kishore-kurri: In most cases, registry should have the key->value but added fallback so that any missing case also update will be done based on cleaned Name

src/migration/NameMappingRegistry.ts
sf-kishore-kurri: will check

src/migration/flexcard.ts
sf-kishore-kurri: Ideally we shouldn't see any misses because registry is prepared based on pre-migrated data and else logic is fallback to avoid not clean up names

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -129,59 +130,44 @@ export default class Migrate extends OmniStudioBaseCommand {
...
     Logger.log(messages.getMessage('apiVersionInfo', [apiVersion]));
     Logger.logVerbose(messages.getMessage('migrationTargets', [migrateOnly || 'all']));
     Logger.logVerbose(messages.getMessage('relatedObjectsInfo', [relatedObjects || 'none']));
     Logger.logVerbose(messages.getMessage('allVersionsFlagInfo', [String(allVersions)]));
 
-    // const includeLwc = this.flags.lwc ? await this.ux.confirm('Do you want to include LWC migration? (yes/no)') : false;
-    // Register the migration objects
+    // Initialize the name mapping registry and pre-process all components
+    const nameRegistry = NameMappingRegistry.getInstance();
+    nameRegistry.clear(); // Clear any previous mappings
```
snehaljha-sf, line 147: We don't need to clear the mappigns as we are not persisting it in any file based storage

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: its existing code and truncate is returning MigratedObject[]

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -129,59 +130,44 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+
+    Logger.log(messages.getMessage('startingComponentPreProcessing'));
+    await this.preProcessAllComponents(nameRegistry, namespace, conn, migrateOnly);
```
snehaljha-sf: nameRegistry follows a singleton pattern, we don't need to add this dependency while calling a function

sf-kishore-kurri: you mean NameMappingRegistry.getInstance() should be good instead of passing

src/commands/omnistudio/migration/migrate.ts
snehaljha-sf: We can change teh declaration visibility or add new method. It's better to have all the queries to be defined at same/similar place

src/migration/NameMappingRegistry.ts
```diff
@@ -0,0 +1,401 @@
...
+ * Centralized registry for tracking name mappings and managing dependency updates
+ * during Omnistudio component migration.
+ */
+export class NameMappingRegistry {
+  private static instance: NameMappingRegistry;
+  private nameMappings: Map<string, ComponentNameMapping> = new Map();
+  private dependencyReferences: DependencyReference[] = [];
+
+  // Type-specific mappings for quick lookup
+  private dataMapperMappings: Map<string, string> = new Map(); // original -> cleaned
```
snehaljha-sf, line 36: why do we need componentWise name mapping if we already have common name mapping with prefix

src/migration/flexcard.ts
snehaljha-sf: We should add some logs to let us know about any such miss

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -240,6 +226,51 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+          process.exit(1);
+        }
+      }
+      // Check for general consent to make modifications with OMT
+      const generalConsent = await this.getGeneralConsent();
+      if (generalConsent) {
+        // Use ProjectPathUtil for APEX project folder selection (matches assess.ts logic)
+        projectPath = await ProjectPathUtil.getProjectPath(messages, true);
+        targetApexNamespace = await this.getTargetApexNamespace(objectsToProcess, targetApexNamespace);
+        await this.handleExperienceSitePrerequisites(
```
sf-aastha-paruthi: This has been moved to premigrate class.

sf-kishore-kurri: yes, its merge conflict issue taking care of it

src/migration/NameMappingRegistry.ts
sf-kishore-kurri: @snehaljha-sf component mapping is for faster lookup

src/migration/flexcard.ts
sf-kishore-kurri: sure @snehaljha-sf

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: The component mapping query only fetching 2-3 columns where as migration need more columns for data copy so we can leave the mapping query separate from data migration ones  @snehaljha-sf

sf-kishore-kurri:
@snehaljha-sf @sf-aastha-paruthi Please review the changes

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: clearing post init

snehaljha-sf (approved):

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -129,59 +130,43 @@ export default class Migrate extends OmniStudioBaseCommand {
...
-    // We need to truncate the standard objects first
-    let objectMigrationResults = await this.truncateObjects(migrationObjects, debugTimer);
-    objectMigrationResults = objectMigrationResults.filter(
-      (result) => result.name !== Constants.GlobalAutoNumberComponentName
-    );
+    // We need to truncate the standard objects first (in reverse order for cleanup)
+    let objectMigrationResults = await this.truncateObjects([...migrationObjects].reverse(), debugTimer);
     const allTruncateComplete = objectMigrationResults.length === 0;
 
     if (allTruncateComplete) {
```
sf-aastha-paruthi, line 168: Please verify by running migration and assessment with only os, fc.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -240,6 +225,59 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+          messages,
+          this.ux
+        );
+        await preMigrate.handleExperienceSitePrerequisites(
+          objectsToProcess,
+          conn,
+          isExperienceBundleMetadataAPIProgramaticallyEnabled
+        );
+        Logger.logVerbose(
+          'The objects to process after handleExpSitePrerequisite are ' + JSON.stringify(objectsToProcess)
```
sf-aastha-paruthi, line 265: This has been moved to message.json

src/commands/omnistudio/migration/assess.ts
```diff
@@ -208,7 +208,8 @@ export default class Assess extends OmniStudioBaseCommand {
       // If no specific component is specified, assess all components
       await this.assessDataRaptors(assesmentInfo, namespace, conn);
       await this.assessFlexCards(assesmentInfo, namespace, conn, allVersions);
-      await this.assessOmniScripts(assesmentInfo, namespace, conn, allVersions, OmniScriptExportType.All);
+      await this.assessOmniScripts(assesmentInfo, namespace, conn, allVersions, OmniScriptExportType.OS);
```
sf-aastha-paruthi, line 213: Shouldn't the order of assessment and migration be same, given that there can be FC->OS and OS->FC dependencies.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -340,18 +380,37 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+  private getMigrationObjectsInCorrectOrder(
     migrateOnly: string,
     migrationObjects: MigrationTool[],
     namespace: string,
-    conn,
+    conn: any,
     allVersions: any
   ): MigrationTool[] {
     if (!migrateOnly) {
+      // Correct order: DataMapper -> OmniScript/IP -> FlexCard -> GlobalAutoNumber
```
sf-aastha-paruthi: nit, remove duplicate comment, comment in line 383 already covers the ordering.

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -406,6 +467,137 @@ export default class Migrate extends OmniStudioBaseCommand {
...
+      }
+    }
+
+    return { lwc, angular };
+  }
+
+  /**
+   * Query FlexCards from the org
+   */
+  private async queryFlexCards(conn: any, namespace: string): Promise<any[]> {
```
sf-aastha-paruthi, line 550: All the 4 functions are making the same call to query and then return the resuts, just the query is different. The query can be separated and the other part can be put to a common function to avoid duplicacy.

src/migration/omniscript.ts
```diff
@@ -196,10 +200,11 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     flexCardAssessmentInfos: FlexCardAssessmentInfo[]
   ): Promise<OmniAssessmentInfo> {
     try {
-      Logger.log(this.messages.getMessage('startingOmniScriptAssessment'));
+      const exportComponentType =
+        this.exportType === OmniScriptExportType.IP ? 'Integration Procedures' : 'Omniscripts';
```
sf-aastha-paruthi: getName() can be used.

src/migration/omniscript.ts
```diff
@@ -231,7 +236,9 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     const existingDataRaptorNames = new Set(dataRaptorAssessmentInfos.map((info) => info.name));
     const existingFlexCardNames = new Set(flexCardAssessmentInfos.map((info) => info.name));
 
-    const progressBar = createProgressBar('Assessing', 'Omniscript and Integration Procedure');
+    const progressBarType: ComponentType =
+      this.exportType === OmniScriptExportType.IP ? 'Integration Procedures' : 'Omniscripts';
```
sf-aastha-paruthi: getName()

sf-aastha-paruthi (approved):

src/migration/omniscript.ts
sf-kishore-kurri: Done

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: Even on latest, we aren't using the messages

sf-kishore-kurri:
@snehaljha-sf @sf-aastha-paruthi addressed the review comments

PR #360, opened 2025-08-08
Function enhancements

mdmehran-qureshi:
Omniscript migration doesn’t take care of reserved/invalid names attached with response transformation names of node/path of DM remote action
Element Name has to be unique for the same OmniScript is identified in migrate mode but not in assessment mode

### DIFF
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -687,6 +687,38 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
 
       // Get All elements for each OmniScript__c record(i.e IP/OS)
       const elements = await this.getAllElementsForOmniScript(recordId);
+
+      // Check for duplicate element names within the same OmniScript
+      const elementNames = new Set<string>();
+      const duplicateElementNames = new Set<string>();
+
+      for (const elem of elements) {
+        const elemName = elem['Name'];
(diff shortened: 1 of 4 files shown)


src/migration/omniscript.ts
```diff
@@ -493,6 +525,12 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       existingOmniscriptNames.add(recordName);
     }
 
+    // Add warning for duplicate element names within the same OmniScript
+    if (duplicateElementNames.size > 0) {
+      const duplicateNamesList = Array.from(duplicateElementNames).join(', ');
```
sf-kishore-kurri, line 524: what will be status when we found the duplicates? and also does it not applicable to Flexcard?

sf-kishore-kurri: when duplicates found it should be tag as Need Manual intervention so that customer can act on it before migration

snehaljha-sf:
Apart from @sf-kishore-kurri  comment lgtm

snehaljha-sf (approved):

src/migration/omniscript.ts
mdmehran-qureshi: Discussed

sf-kishore-kurri (approved):
Looks good to me

PR #361, opened 2025-08-08
feat: @W-18994070 : Naming convention and clean of flags on assess an…

sf-kishore-kurri:
…d migrate mode

Changes:
- Clean up of apiversion, namespace and json flags
- Error handling
- README udpate and clean of removed flags


snehaljha-sf (approved):

src/commands/omnistudio/migration/assess.ts
```diff
@@ -27,10 +27,6 @@ export default class Assess extends OmniStudioBaseCommand {
   public static args = [{ name: 'file' }];
 
   protected static flagsConfig = {
```
sf-aastha-paruthi, line 29: Is the namespace flag being removed and are we going with showing the packages and then user selection?

sf-aastha-paruthi (approved):

src/commands/omnistudio/migration/assess.ts
sf-kishore-kurri: yes and even if we provide namespace flag we aren't considering

PR #366, opened 2025-08-11
feat: @W-18480479 : Enabling standard runtime along with data model

sf-kishore-kurri:
Changes:
- Clean up existing loigc and added OmnistudioSettings pref manager
- Handling autonumber and std runtime
- added the tests


snehaljha-sf (approved):

src/migration/postMigrate.ts
```diff
@@ -43,26 +85,90 @@ export class PostMigrate extends BaseMigrationTool {
   ): Promise<string[]> {
     try {
       Logger.logVerbose(this.messages.getMessage('settingDesignersToStandardModel'));
+
+      // First check if standard designer is already enabled for this package
+      Logger.logVerbose(this.messages.getMessage('checkingStandardDesignerStatus', [namespaceToModify]));
```
sf-aastha-paruthi, line 93: This function has been modified to do multiple things. Is should only deal with standard data model. Please create new functions for other tasks and combine all those in one separate function if needed.

src/utils/OmnistudioSettingsPrefManager.ts
```diff
@@ -0,0 +1,81 @@
+import { Connection, Messages } from '@salesforce/core';
+import { Logger } from './logger';
+import { MetadataInfo } from './interfaces';
+
+export class OmnistudioSettingsPrefManager {
```
sf-aastha-paruthi, line 5: There is already a class OrgPreferences having the same responsibility. Please move these functions to that.

src/migration/postMigrate.ts
sf-kishore-kurri: Now the function is only doing the data model switch and then invoke runtime switch

sf-aastha-paruthi (changes requested):

src/migration/postMigrate.ts
sf-aastha-paruthi: Within runtime switch its doing some prechecks and all. This can all be taken in separate function and each can have single responsibility.

src/migration/postMigrate.ts
```diff
@@ -35,32 +37,123 @@ export class PostMigrate extends BaseMigrationTool {
...
         `;
 
       const result: ExecuteAnonymousResult = await AnonymousApexRunner.run(this.org, apexCode);
 
       if (result?.success === false) {
-        const message = result?.exceptionStackTrace;
+        const message = result?.exceptionStackTrace || result?.exceptionMessage || 'Unknown error';
+
         Logger.error(this.messages.getMessage('errorSettingDesignersToStandardModel', [message]));
+        Logger.logVerbose(this.messages.getMessage('skipStandardRuntimeDueToFailure'));
```
sf-aastha-paruthi, line 111: Shouldn't the message be data model related.

sf-aastha-paruthi (approved):

src/migration/postMigrate.ts
sf-kishore-kurri: @sf-aastha-paruthi When the data model errored out, we are also adding message to skip standard runtime

PR #387, opened 2025-08-19
fix: skip Angular OmniScripts from duplicate name check in assessment…

mdmehran-qureshi:
… phase

- Added early detection of Angular OmniScripts before duplicate name validation
- Angular OmniScripts are now excluded from existingOmniscriptNames set
- Prevents Angular OmniScripts from being compared for duplicates since they won't be migrated
- Maintains proper assessment flow: skip Angular first, then handle duplicate names


src/migration/omniscript.ts
```diff
@@ -466,6 +466,15 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       ? 'Integration Procedure'
       : 'OmniScript';
 
+    // Check if this is an Angular OmniScript first - if so, skip duplicate name checks
+    let isAngularOmniScript = false;
+    if (omniProcessType === this.OMNISCRIPT) {
+      const type = omniscript[this.namespacePrefix + 'IsLwcEnabled__c'] ? 'LWC' : 'Angular';
+      if (type === 'Angular') {
```
sf-kishore-kurri, line 473: Can we skip the remaining steps for Angular case instead of doing explicit checks

PR #388, opened 2025-08-19
fix: manual step call to action

snehaljha-sf:


sf-kishore-kurri (approved):
Approving the changes

PR #390, opened 2025-08-22
feat: @W-18480496 : Update reference fixes for IP and Omniscript

sf-kishore-kurri:
Changes:
- Omniscript or Integration Procedures uses OmniProcessCompilation definition with runtime
- Migration Assistant doesn't update compilation causing issues with runtime
- Code fix to update the references in compilation definition
- Added test cases for regression testing

### DIFF
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -1753,6 +1753,12 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
         propSetMap.integrationProcedureKey = this.nameRegistry.getIntegrationProcedureCleanedName(key);
       } else {
         const parts = key.split('_');
+        // Integration Procedures should have Type_SubType format (2 parts)
+        if (parts.length > 2) {
+          Logger.logVerbose(this.messages.getMessage('integrationProcedureInvalidUnderscoreFormat', [key]));
+          return;
+        }
+
         propSetMap.integrationProcedureKey = parts.map((p) => this.cleanName(p, true)).join('_');
       }
     }
@@ -1787,9 +1793,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
 
     if (this.nameRegistry.isAngularOmniScript(fullOmniScriptName)) {
       // Keep original reference as-is since Angular OmniScript won't be migrated
-      propSetMap['Type'] = osType;
-      propSetMap['Sub Type'] = osSubType;
-      propSetMap['Language'] = osLanguage;
+      return;
     } else if (this.nameRegistry.hasOmniScriptMapping(fullOmniScriptName)) {
       // Registry has mapping for this LWC OmniScript - extract cleaned parts
       const cleanedFullName = this.nameRegistry.getCleanedName(fullOmniScriptName, 'OmniScript');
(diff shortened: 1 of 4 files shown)


src/migration/omniscript.ts
```diff
@@ -1624,6 +1638,215 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+   */
+  private processStepAction(propSetMap: any): void {
+    // Handle remoteClass - typically references Apex classes
+    if (propSetMap.remoteClass && propSetMap.remoteClass.trim()) {
+      propSetMap.remoteClass = this.cleanName(propSetMap.remoteClass);
+    }
+
+    // Handle remoteMethod - typically references Apex methods
+    if (propSetMap.remoteMethod && propSetMap.remoteMethod.trim()) {
+      propSetMap.remoteMethod = this.cleanName(propSetMap.remoteMethod);
```
snehaljha-sf: Are we cleaning class name and method name as well as part of migration?

src/migration/omniscript.ts
```diff
@@ -1624,6 +1638,215 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+    }
+
+    // Handle remoteOptions pre/post transform bundles if they exist in Step elements
+    if (propSetMap.remoteOptions) {
+      if (propSetMap.remoteOptions.preTransformBundle) {
+        const bundleName = propSetMap.remoteOptions.preTransformBundle;
+        if (this.nameRegistry.hasDataMapperMapping(bundleName)) {
+          propSetMap.remoteOptions.preTransformBundle = this.nameRegistry.getDataMapperCleanedName(bundleName);
+        } else {
+          propSetMap.remoteOptions.preTransformBundle = this.cleanName(bundleName);
```
snehaljha-sf, line 1830: in which case we'll get in else block

src/migration/omniscript.ts
```diff
@@ -1624,6 +1638,215 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+    const osType = propSetMap['Type'] || '';
+    const osSubType = propSetMap['Sub Type'] || '';
+    const osLanguage = propSetMap['Language'] || 'English';
+
+    // Construct full OmniScript name to check registry
+    const fullOmniScriptName = `${osType}_${osSubType}_${osLanguage}`;
+
+    if (this.nameRegistry.isAngularOmniScript(fullOmniScriptName)) {
+      // Keep original reference as-is since Angular OmniScript won't be migrated
+      propSetMap['Type'] = osType;
```
snehaljha-sf: why do we need to set the values again?

src/migration/omniscript.ts
```diff
@@ -1624,6 +1638,215 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+      propSetMap['Sub Type'] = osSubType;
+      propSetMap['Language'] = osLanguage;
+    } else if (this.nameRegistry.hasOmniScriptMapping(fullOmniScriptName)) {
+      // Registry has mapping for this LWC OmniScript - extract cleaned parts
+      const cleanedFullName = this.nameRegistry.getCleanedName(fullOmniScriptName, 'OmniScript');
```
snehaljha-sf, line 1799: While preparing how are we handling OS names which contain multiple _

src/migration/omniscript.ts
```diff
@@ -1624,6 +1638,215 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+    }
+
+    // Handle integrationProcedureKey
+    if (propSetMap.integrationProcedureKey) {
+      const key = propSetMap.integrationProcedureKey;
+      if (this.nameRegistry.hasIntegrationProcedureMapping(key)) {
+        propSetMap.integrationProcedureKey = this.nameRegistry.getIntegrationProcedureCleanedName(key);
+      } else {
+        const parts = key.split('_');
+        propSetMap.integrationProcedureKey = parts.map((p) => this.cleanName(p, true)).join('_');
```
snehaljha-sf, line 1762: We need handle multiple _ for IP type and subtype

snehaljha-sf (commented):
We should add warnings and logs as well for reference updated for these actions

src/migration/omniscript.ts
sf-kishore-kurri: I will clean up the logic, thanks

src/migration/omniscript.ts
sf-kishore-kurri: We clean type/subType and then append them using "_" delimiter

src/migration/omniscript.ts
sf-kishore-kurri: The else block is implemented for fallback

src/migration/omniscript.ts
sf-kishore-kurri: its fallback

snehaljha-sf (approved):

PR #391, opened 2025-08-26
@W-19370170 Flexcard for community pages

sf-aastha-paruthi:
### What does this PR do?
Flexcards to be enabled for community pages for migration
Tests cases for flexcards

### DIFF
--- a/src/migration/flexcard.ts
+++ b/src/migration/flexcard.ts
@@ -89,6 +89,7 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
   async migrate(): Promise<MigrationResult[]> {
     // Get All the Active VlocityCard__c records
     const allCards = await this.getAllActiveCards();
+
     Logger.log(this.messages.getMessage('foundFlexCardsToMigrate', [allCards.length]));
 
     // Filter out FlexCards with Angular OmniScript dependencies
@@ -1041,6 +1042,8 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
       mappedObject[CardMappings.Datasource__c] = JSON.stringify(datasource);
     }
 
+    this.ensureCommunityTargets(mappedObject);
+
     // Update all dependencies comprehensively
     this.updateAllDependenciesWithRegistry(mappedObject, invalidIpNames);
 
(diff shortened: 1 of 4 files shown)


src/migration/flexcard.ts
```diff
@@ -1532,4 +1535,42 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+      definition.xmlObject.targets = { target: [] };
+    }
+
+    // Ensure target is an array
+    // This also does not seem to be correct and may not be needed as the specified json structure would exist already
+    if (!Array.isArray(definition.xmlObject.targets.target)) {
+      definition.xmlObject.targets.target = [definition.xmlObject.targets.target];
+    }
+
+    const requiredTargets = ['lightningCommunity__Page', 'lightningCommunity__Default'];
```
sf-kishore-kurri: @sf-aastha-paruthi  Can you add other targets also like [record page, app and home page] to handle the child flexcard case

src/migration/flexcard.ts
```diff
@@ -1532,4 +1535,42 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+    const definition = JSON.parse(mappedObject[CardMappings.Definition__c] || '{}');
+
+    if (!definition || !definition.xmlObject) {
+      return;
+    }
+
+    // Initialize targets structure if it doesn't exist
+    if (!definition.xmlObject.targets) {
+      definition.xmlObject.targets = { target: [] };
+    }
```
snehaljha-sf, line 1558: if targets is defiend but targets.target is not the constructed value of targets will contain [undefined.......] Shall we handle this or this case won't exist?

snehaljha-sf (commented):
Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them

sf-kishore-kurri:
> Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them

@sf-aastha-paruthi Do we have way to check whether flexcard is already published or not?

sf-aastha-paruthi:
> targets

@sf-kishore-kurri , this PR is as per the discussion we had. There is no direct way to identify the flexcards used in community pages. So we agreed upon enabling the setting for all the flexcards.

sf-kishore-kurri:
> > Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them
> 
> @sf-aastha-paruthi Do we have way to check whether flexcard is already published or not?

Yes, but can we segregate published vs non published ones  based on flexcard definition?

sf-aastha-paruthi:
> > > Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them
> > 
> > 
> > @sf-aastha-paruthi Do we have way to check whether flexcard is already published or not?
> 
> Yes, but can we segregate published vs non published ones based on flexcard definition?

No, we can't

sf-kishore-kurri:
> > > > Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them
> > > 
> > > 
> > > @sf-aastha-paruthi Do we have way to check whether flexcard is already published or not?
> > 
> > 
> > Yes, but can we segregate published vs non published ones based on flexcard definition?
> 
> No, we can't

@sf-aastha-paruthi I checked the flexcard definition so we can add one more check whether FC is acitve or not? If it is active  only add the additional targets. When the card is not active publish action is disabled in UI

Screenshot 2025-09-01 at 5 58 38 PM

src/migration/flexcard.ts
sf-aastha-paruthi: Updated

sf-aastha-paruthi:
> > > > > Aren't we suppose to handle only those flexcard which is used in expsites. Here we're adding targets in all of them
> > > > 
> > > > 
> > > > @sf-aastha-paruthi Do we have way to check whether flexcard is already published or not?
> > > 
> > > 
> > > Yes, but can we segregate published vs non published ones based on flexcard definition?
> > 
> > 
> > No, we can't
> 
> @sf-aastha-paruthi I checked the flexcard definition so we can add one more check whether FC is acitve or not? If it is active only add the additional targets. When the card is not active publish action is disabled in UI
> 
> Screenshot 2025-09-01 at 5 58 38 PM

Done

sf-kishore-kurri (approved):
@sf-aastha-paruthi Approving the changes, but Let's not merge the changes until we hear from experience cloud team for the perf/scale impact

sf-aastha-paruthi:
As got the confirmation from Krishna, going ahead with the merge.

snehaljha-sf (approved):

PR #393, opened 2025-08-28
feat: @W-19268842 : Feedback and filter pop up issue

sf-kishore-kurri:


src/commands/omnistudio/migration/assess.ts
```diff
@@ -155,6 +156,12 @@ export default class Assess extends OmniStudioBaseCommand {
       Logger.log((error as Error).stack);
     }
     await AssessmentReporter.generate(assesmentInfo, conn.instanceUrl, orgs, assessOnly, objectsToProcess, messages);
+    Logger.log(
+      messages.getMessage('assessmentSuccessfulMessage', [
+        orgs.orgDetails?.Id,
+        path.join(process.cwd(), 'assessment_reports'),
```
snehaljha-sf: Since we're using the dir name in multiple files now, Let's move it to StringConstants and use the same in both files

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -198,7 +199,12 @@ export default class Migrate extends OmniStudioBaseCommand {
       actionItems,
       objectsToProcess
     );
-
+    Logger.log(
+      messages.getMessage('migrationSuccessfulMessage', [
+        orgs.orgDetails?.Id,
+        path.join(process.cwd(), 'migration_report'),
```
snehaljha-sf: same here

src/javascripts/reportGeneratorUtility.js
```diff
@@ -11,9 +11,35 @@ function toggleFilterDropdown(tableId) {
   }
 }
 
-function toggleDiffModal(name) {
```
snehaljha-sf, line 14: Was this defined 2 times earlier?

snehaljha-sf (commented):
Added comments

src/javascripts/reportGeneratorUtility.js
sf-kishore-kurri: yes

snehaljha-sf (approved):

PR #413, opened 2025-09-12
Assessment support for standard entities

sf-aastha-paruthi:

### DIFF
--- a/src/commands/omnistudio/migration/migrate.ts
+++ b/src/commands/omnistudio/migration/migrate.ts
@@ -33,6 +33,7 @@ import { PreMigrate } from '../../../migration/premigrate';
 import { GlobalAutoNumberMigrationTool } from '../../../migration/globalautonumber';
 import { ValidatorService } from '../../../utils/validatorService';
 import { NameMappingRegistry } from '../../../migration/NameMappingRegistry';
+import { ISUSECASE2 } from '../../../utils/constants/migrationConfig';
 
 // Initialize Messages with the current plugin directory
 Messages.importMessagesDirectory(__dirname);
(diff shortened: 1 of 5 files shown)


src/commands/omnistudio/migration/migrate.ts
```diff
@@ -33,6 +33,7 @@ import { PreMigrate } from '../../../migration/premigrate';
 import { GlobalAutoNumberMigrationTool } from '../../../migration/globalautonumber';
 import { ValidatorService } from '../../../utils/validatorService';
 import { NameMappingRegistry } from '../../../migration/NameMappingRegistry';
+import { ISUSECASE2 } from '../../../utils/constants/migrationConfig';
```
sf-kishore-kurri, line 36: Can you rename this to standardDataModel instead of usecase2, so that we can use the same for use case 3 as well

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -339,6 +338,10 @@ export default class Migrate extends OmniStudioBaseCommand {
   }
 
   private async truncateObjects(migrationObjects: MigrationTool[], debugTimer: DebugTimer): Promise<MigratedObject[]> {
+    if (ISUSECASE2) {
+      return;
```
sf-kishore-kurri, line 342: for std data model, can we empty promise so that we don't have to clean up the existing logic

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -172,14 +173,12 @@ export default class Migrate extends OmniStudioBaseCommand {
     const allTruncateComplete = objectMigrationResults.length === 0;
 
     // Log truncation errors if any exist
-    if (!allTruncateComplete) {
+    if (!ISUSECASE2 && !allTruncateComplete) {
```
sf-kishore-kurri, line 176: When you return EMPTY promise then we don't need this change

src/migration/dataraptor.ts
```diff
@@ -23,6 +23,7 @@ import {
 import { StringVal } from '../utils/StringValue/stringval';
 import { Logger } from '../utils/logger';
 import { createProgressBar } from './base';
+import { ISUSECASE2 } from '../utils/constants/migrationConfig';
```
sf-kishore-kurri, line 26: change here as well

src/commands/omnistudio/migration/migrate.ts
sf-kishore-kurri: As we don't truncate and update the value, can you check whether metadata tables are populated properly with name update

src/migration/omniscript.ts
```diff
@@ -1886,6 +1890,22 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     }
   }
 
+  private getFieldKey(fieldName: string): string {
+    return ISUSECASE2 ? OmniScriptMappings[fieldName] : this.namespacePrefix + fieldName;
+  }
+
+  private getQueryNamespace(): string {
+    return ISUSECASE2 ? '' : this.namespace;
```
sf-kishore-kurri, line 1898: change to std data model

sf-kishore-kurri (commented):
Added comments

PR #414, opened 2025-09-15
fix: @W-19637353: Prudential Migration issues

sf-kishore-kurri:
Changes:
- Hanlded the error case for DM data migration issues
- IP having custom formula causing the JSON parse issue
- Hanlde of LWC component with nested namespace tags

### DIFF
--- a/src/migration/dataraptor.ts
+++ b/src/migration/dataraptor.ts
@@ -175,8 +175,6 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
         continue;
       }
 
-      duplicatedNames.add(transformedDataRaptor['Name']);
-
       // Create a map of the original records
       originalDrRecords.set(recordId, dr);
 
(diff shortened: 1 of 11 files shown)


snehaljha-sf (approved):

src/migration/dataraptor.ts
```diff
@@ -191,21 +189,28 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
...
-        drUploadResponse.newName = transformedDataRaptor[DRBundleMappings.Name];
+        const items = await this.getItemsForDataRaptor(dataRaptorItemsData, name, drUploadResponse.id);
+        drUploadResponse.newName = dataMapperName;
 
         // Move the items
         await this.uploadTransformedData(DataRaptorMigrationTool.OMNIDATATRANSFORMITEM_NAME, items);
       } else {
         // Handle failed migration - add error information
-        if (!drUploadResponse) {
+        if (!drUploadResponse?.success) {
```
sf-aastha-paruthi, line 203: Why are we removing the undefined check?

src/migration/omniscript.ts
```diff
@@ -823,19 +823,75 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+                    // elementValueMap is already an object, use directly
+                    elementValueMap = propertySet.elementValueMap;
+                    Logger.logVerbose(
+                      `ElementValueMap accessed as object for element: ${ipElement['Name'] || 'Unknown'}`
+                    );
+                  } else {
+                    Logger.warn(
+                      this.messages.getMessage('elementValueMapUnexpectedType', [
+                        typeof propertySet.elementValueMap,
+                        ipElement['Name'] || 'Unknown',
```
sf-aastha-paruthi, line 842: Should the code not return from here, if the elementValueMap is of unexpectedType?

src/migration/omniscript.ts
```diff
@@ -823,19 +823,75 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+                            this.namespacePrefix,
+                            value,
+                            functionDefinitionMetadata
+                          );
+                          if (updatedFormula !== value) {
+                            elementValueMap[key] = updatedFormula;
+                            formulasUpdated = true;
+                            Logger.logVerbose(`Updated formula in ${key}: ${value} -> ${updatedFormula}`);
+                          }
+                        } catch (formulaEx) {
```
sf-aastha-paruthi, line 865: In case of exception/ failure to update the formula, how will the same be reflected in the report, if we are catching the exceptions

src/migration/dataraptor.ts
sf-kishore-kurri: We aren't removing, we are checking the success status and logging error for failure case

src/migration/omniscript.ts
sf-kishore-kurri: If elementValueType is undefined then we don't update the config and move on with existing

src/migration/omniscript.ts
sf-kishore-kurri: When update formula function fails, we are logging error and verbose message for debugging and managed package runtime will do the formula migration again post migration so we should be good

sf-aastha-paruthi (approved):

src/migration/omniscript.ts
sf-aastha-paruthi: If the managed package will do it and we are sure, then why are we duplicating the logic?

src/migration/omniscript.ts
sf-aastha-paruthi: If this condition is false
typeof propertySet.elementValueMap === 'object') still the line 847 will be hit
I meant elementValueMapUnexpectedType.

src/migration/omniscript.ts
sf-kishore-kurri: It is the existing logic with OMT tool

PR #420, opened 2025-09-23
@W-19464592 Standard data model changes for name registry

sf-aastha-paruthi:

### DIFF
--- a/src/commands/omnistudio/migration/migrate.ts
+++ b/src/commands/omnistudio/migration/migrate.ts
@@ -613,7 +613,7 @@ export default class Migrate extends OmniStudioBaseCommand {
   private async queryFlexCards(conn: any, namespace: string): Promise<any[]> {
     let query;
     if (isStandardDataModel()) {
-      query = "SELECT Id, Name FROM OmniUiCard WHERE IsActive = true";
+      query = 'SELECT Id, Name FROM OmniUiCard WHERE IsActive = true';
     } else {
       query = `SELECT Id, Name FROM ${namespace}__VlocityCard__c WHERE ${namespace}__CardType__c = 'flex' AND ${namespace}__Active__c = true`;
     }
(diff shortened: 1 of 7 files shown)


src/commands/omnistudio/migration/migrate.ts
```diff
@@ -554,7 +555,14 @@ export default class Migrate extends OmniStudioBaseCommand {
    * Query DataMappers from the org
    */
   private async queryDataMappers(conn: any, namespace: string): Promise<any[]> {
-    const query = `SELECT Id, Name FROM ${namespace}__DRBundle__c WHERE ${namespace}__Type__c != 'Migration'`;
+    let query;
+
+    if (isStandardDataModel()) {
+      query = "SELECT Id, Name FROM OmniDataTransform WHERE Type != 'Migration'";
```
sf-kishore-kurri, line 564: Can we use same pattern of lookup entity name and field names based on data model instead of conditional block?

src/commands/omnistudio/migration/migrate.ts
```diff
@@ -712,4 +732,9 @@ export default class Migrate extends OmniStudioBaseCommand {
       }
     });
   }
+
+  private getFieldKeyForOmniscript(namespacePrefix: string, fieldName: string): string {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-member-access
```
sf-kishore-kurri: Do we have similar API in assess as well?

sf-kishore-kurri (commented):
Added comments, other changes are looks good to me

src/commands/omnistudio/migration/migrate.ts
sf-aastha-paruthi: The current implementation of having separate queries is much cleaner.

shaurabh-tiwari-git (approved):
LGTM

sf-kishore-kurri (approved):
LGTM

PR #428, opened 2025-10-08
fix: @W-19837693: Handling multiple versions migration and assessment for Omniscript/Flexcards

sf-kishore-kurri:
Changes:
- Handle of multiple versions having same name pre/post clean up with allVersions
- Added logic to check for unique and duplicate based on earlier versions
- Added test cases


src/migration/flexcard.ts
```diff
@@ -783,8 +802,16 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
         ? `${transformedCard['Name']}_${transformedCard['VersionNumber']}`
         : transformedCard['Name'];
 
-      if (uniqueNames.has(uniqueCheckName)) {
-        this.setRecordErrors(card, this.messages.getMessage('duplicatedCardName', [uniqueCheckName]));
+      if (uniqueNames.has(uniqueCheckName) || dupFlexCardNames.has(transformedCard['Name'])) {
```
sf-aastha-paruthi: The changes in this file appear similar. Please move to common function.

src/migration/flexcard.ts
```diff
@@ -230,11 +235,23 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
-    if (uniqueNames.has(uniqueCleanedName)) {
-      flexCardAssessmentInfo.warnings.push(this.messages.getMessage('duplicateCardNameMessage', [uniqueCleanedName]));
+    if (uniqueNames.has(uniqueCleanedName) || dupFlexCardNames.has(originalName)) {
+      // Add the original flexcard name to the set
+      if (this.allVersions && !dupFlexCardNames.has(originalName)) {
+        dupFlexCardNames.add(originalName);
+      }
+      if (uniqueNames.has(uniqueCleanedName)) {
+        flexCardAssessmentInfo.warnings.push(this.messages.getMessage('duplicateCardNameMessage', [uniqueCleanedName]));
+      } else {
```
sf-aastha-paruthi: The if or else will be set to true when multiple versions of flexcards are there.
A - v1
A - v2
So warning will get populated which is not intended.

sf-kishore-kurri: Discussed with @sf-aastha-paruthi and resolved the scenario, but FC delete case is missing

src/migration/flexcard.ts
```diff
@@ -231,11 +236,29 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
     // Check for duplicate names (include version when allVersions is true)
     const uniqueCleanedName = this.allVersions ? `${cleanedName}_${version}` : cleanedName;
+
+    // Check for exact duplicate (same name + same version)
     if (uniqueNames.has(uniqueCleanedName)) {
       flexCardAssessmentInfo.warnings.push(this.messages.getMessage('duplicateCardNameMessage', [uniqueCleanedName]));
       assessmentStatus = getUpdatedAssessmentStatus(assessmentStatus, 'Needs manual intervention');
     }
+    // Check for naming conflict: different original names cleaning to same name
+    else if (this.allVersions && dupFlexCardNames.has(cleanedName)) {
```
sf-aastha-paruthi, line 246: The code is duplicate. Can be separated to a separate function.

test/migration/omniscript-allversions-duplicate.test.ts
```diff
@@ -314,12 +314,230 @@ describe('OmniScript Multiple Versions Duplicate Check', () => {
...
+
+      expect(resultA2.migrationStatus).to.equal('Ready for migration');
+      expect(resultA2.name).to.equal('A_Type_English_2');
+      expect(resultA2.warnings).to.have.length(0);
+
+      // A$_3 should be flagged as lower version duplicate (different original name cleaning to A)
+      expect(resultAS3.migrationStatus).to.equal('Needs manual intervention');
+      expect(resultAS3.name).to.equal('A_Type_English_3');
+      expect(resultAS3.warnings.length).to.be.greaterThan(0);
+      expect(resultAS3.warnings.some((w) => w.includes('Lower version duplicate'))).to.be.true;
```
sf-aastha-paruthi, line 533: Why is it being populated with lower version duplicate issue?

sf-kishore-kurri: We don't have exact duplicate exists so bucketed under lower version duplicate case, if we had version1/2 which had marked for duplicate

src/migration/flexcard.ts
sf-kishore-kurri: with migration, we set the status on record whereas in assessment its only message with status

test/migration/omniscript-allversions-duplicate.test.ts
sf-aastha-paruthi: A %s with name \"%s\" will not be migrated because lower version of same %s will be marked as duplicate, which could lead to conflicts during migration.

But the message gives the impression that, the omniscript has some lower version, which has been marked as duplicate.

Can you please elaborate what the message trying to convey?

PR #429, opened 2025-10-09
@W-19821865 Flexcard Name update for standard model

sf-aastha-paruthi:
### What does this PR do?
Earlier Core side of flexcard did not allow name update, so the OMT report showed this in the Manual Intervention Category.

Now changes have been made in core side to allow name update. So OMT flexcard has been updated to automate the same.

### DIFF
--- a/src/migration/flexcard.ts
+++ b/src/migration/flexcard.ts
@@ -1103,10 +1103,8 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
       mappedObject[CardMappings.Datasource__c] = JSON.stringify(datasource);
     }
 
-    if (!this.IS_STANDARD_DATA_MODEL) {
-      const isCardActive: boolean = cardRecord[this.getFieldKey('Active__c')];
-      this.ensureCommunityTargets(mappedObject, isCardActive);
-    }
+    const isCardActive: boolean = cardRecord[this.getFieldKey('Active__c')];
+    this.ensureCommunityTargets(mappedObject, isCardActive);
 
     // Update all dependencies comprehensively
     this.updateAllDependenciesWithRegistry(mappedObject, invalidIpNames);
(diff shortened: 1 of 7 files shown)


shaurabh-tiwari-git (approved):
LGTM

src/migration/flexcard.ts
```diff
@@ -89,8 +89,7 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
 
   // Perform Records Migration from VlocityCard__c to OmniUiCard
   async migrate(): Promise<MigrationResult[]> {
-    // Get All the Active VlocityCard__c records
-    const allCards = await this.getAllActiveCards();
+    const allCards = await this.getAllCards();
```
sf-kishore-kurri, line 92: why it is changed from Active -> AllCards here? @sf-aastha-paruthi

sf-aastha-paruthi: Because if allVersions is passed, the function gets active, deactive cards.
The getAllActiveCards is bit misleading.

src/migration/flexcard.ts
```diff
@@ -1146,8 +1103,10 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
       mappedObject[CardMappings.Datasource__c] = JSON.stringify(datasource);
     }
 
-    const isCardActive: boolean = cardRecord[`${this.namespacePrefix}Active__c`];
-    this.ensureCommunityTargets(mappedObject, isCardActive);
+    if (!this.IS_STANDARD_DATA_MODEL) {
+      const isCardActive: boolean = cardRecord[this.getFieldKey('Active__c')];
+      this.ensureCommunityTargets(mappedObject, isCardActive);
```
sf-kishore-kurri: @sf-aastha-paruthi Don't we need target publish logic with std data model also? If they have used the LWC using LWC Omni wrapper?

sf-kishore-kurri (commented):
Added comment for target publish for FC, remaining changes looks good to me

src/migration/flexcard.ts
sf-aastha-paruthi: @sf-kishore-kurri ,
We were doing it for usecase1 because when we were going from custom data model to standard data model, the custom did not need to be explicity published but were still getting rendered on experience sites.
For usecase2 the data model will not change, so we dont need to make the changes.

sf-kishore-kurri: @sf-aastha-paruthi If I remember correctly, Flexcard without publish to exp sites we can use it as LWC using vlocityLWC wrapper, but when we replace with Flexcard std wrapper it should be published so we added this logic

sf-aastha-paruthi: Made the changes

sf-kishore-kurri (approved):
LGTM

PR #441, opened 2025-11-03
@W-19821896 Release workflow update

sf-aastha-paruthi:


.github/workflows/create-github-release.yml
```diff
@@ -5,6 +5,7 @@ on:
     branches:
       - main
       - prerelease/**
+      - release/**
```
snehaljha-sf: This will not match with our branch, we have to use release*

snehaljha-sf (approved):

PR #442, opened 2025-11-07
Blitz fixes - P1 issues

sf-aastha-paruthi:
### What does this PR do?

Apex assessment+Deployment fails | Apex assessment adds duplicate System.callable when the Apex already implements System.callableWhen Apex classes has two System.callable interfaces because of above issue

Flexcard migration fails for original records even though assessment shows can be migratedFlexipage | Flexcard migration fails for naming change records. This happens only if the assessment report has Duplicate Markings for records whose name is clashing and they are not resolved before migration

### DIFF
--- a/src/utils/recordPrioritization.ts
+++ b/src/utils/recordPrioritization.ts
@@ -1,0 +1,72 @@
+/**
+ * Utility functions for record prioritization based on name characteristics
+ */
+
+import { AnyJson } from '@salesforce/ts-types';
+
+/**
+ * Checks if a string is a valid alphanumeric name for migration:
+ * - Must start with a letter (a-z, A-Z)
+ * - Can contain letters and numbers (a-z, A-Z, 0-9) after the first character
+ * - Cannot start with a number
+ * - Cannot contain special characters
+ *
+ * @param str - The string to check
+ * @returns true if the string is valid (starts with letter, alphanumeric only), false otherwise
+ *
+ * @example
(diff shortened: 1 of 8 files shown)


src/utils/recordPrioritization.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @returns true if the string is valid (starts with letter, alphanumeric only), false otherwise
+ *
+ * @example
+ * hasOnlyAlphanumericCharacters('MyName123') // true
+ * hasOnlyAlphanumericCharacters('Test1') // true
+ * hasOnlyAlphanumericCharacters('123Test') // false - starts with number
+ * hasOnlyAlphanumericCharacters('My-Name') // false - contains special character
+ */
+export function hasOnlyAlphanumericCharacters(str: string): boolean {
+  // Must start with a letter, followed by any number of alphanumeric characters
```
sf-kishore-kurri, line 24: We can also create records starting with number instead of letter, might fail this scenario @sf-aastha-paruthi

sf-kishore-kurri (approved):
Approving the changes, but we need to test changes with different alpha-numeric combinations like starting with numbers, underscore etc

src/utils/recordPrioritization.ts
sf-aastha-paruthi: This needs discussion as metadata does not allow such records.

PR #449, opened 2025-11-19
Global Auto Number report update and version name update

sf-aastha-paruthi:
### What does this PR do?
image

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/plugin-omnistudio-migration-tool",
   "description": "This SFDX plugin migrates FlexCard, OmniScript, DataRaptor, and Integration Procedure custom objects to standard objects.",
-  "version": "2.0.0-dev.9",
+  "version": "2.0.0-dev.9-preview",
   "author": "Salesforce",
   "bugs": "https://github.com/forcedotcom/cli/issues",
   "dependencies": {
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@salesforce/plugin-omnistudio-migration-tool",
   "description": "This SFDX plugin migrates FlexCard, OmniScript, DataRaptor, and Integration Procedure custom objects to standard objects.",
-  "version": "2.0.0-dev.9",
+  "version": "2.0.0-dev.9-preview",
```
sf-kishore-kurri, line 4: Did you check whether the version format is good from plugin publish perspective?

src/utils/reportGenerator/reportInterfaces.ts
```diff
@@ -16,6 +16,7 @@ export interface SummaryItemParam {
   total: number;
   data: SummaryItemDetailParam[];
   file: string;
+  showDetails?: boolean; // Set to false to hide counts and view report button
```
sf-kishore-kurri, line 19: can we set default value as true so that we can set false for autonumbber

sf-kishore-kurri (approved):
Approving with comments

src/utils/reportGenerator/reportInterfaces.ts
sf-aastha-paruthi: This is an optional value and if showDetails is false, the Total Records and other details shown in the dashboard card would not be shown. In the html the condition is checking for false.

package.json
sf-aastha-paruthi: Yes, have checked and will give it a try post the plugin publish.
Naming is valid as such.

PR #455, opened 2025-12-10
@W-20469531 In Assessment Report - For Experience Pages, difference in changes column is not getting populated

sf-aastha-paruthi:
### What does this PR do?
In Assessment Report - For Experience Pages, difference in changes column is not getting populated

### DIFF
--- a/src/migration/related/ExperienceSiteMigration.ts
+++ b/src/migration/related/ExperienceSiteMigration.ts
@@ -178,7 +178,12 @@ export class ExperienceSiteMigration extends BaseRelatedObjectMigration {
     Logger.logVerbose(this.messages.getMessage('printDifference', [JSON.stringify(difference)]));
 
     // If there are no differences, mark as not having OmniStudio content to exclude from report
-    if (difference.length === 0) {
+    // Only exclude if status is 'Ready for migration' or 'Successfully migrated' (no warnings/errors)
+    if (
+      difference.length === 0 &&
+      (experienceSiteAssessmentInfo.status === 'Ready for migration' ||
+        experienceSiteAssessmentInfo.status === 'Successfully migrated')
+    ) {
       experienceSiteAssessmentInfo.hasOmnistudioContentWithChanges = false;
       return experienceSiteAssessmentInfo;
     }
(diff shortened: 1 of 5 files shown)


src/migration/related/ExperienceSiteMigration.ts
```diff
@@ -103,7 +103,7 @@ export class ExperienceSiteMigration extends BaseRelatedObjectMigration {
         }
         try {
           const experienceSitePageInfo = this.processExperienceSite(file, type);
-          if (experienceSitePageInfo?.hasOmnistudioContent === true) {
+          if (experienceSitePageInfo?.hasOmnistudioContentWithChanges === true) {
```
sf-kishore-kurri, line 106: @sf-aastha-paruthi Do we need similar change for flexipages also?

sf-kishore-kurri (approved):
@sf-aastha-paruthi check once whether we need similar change for flexipages or not? Changes are looking good

src/migration/related/ExperienceSiteMigration.ts
sf-aastha-paruthi: Done.

src/migration/related/ExperienceSiteMigration.ts
```diff
@@ -177,6 +177,17 @@ export class ExperienceSiteMigration extends BaseRelatedObjectMigration {
 
     Logger.logVerbose(this.messages.getMessage('printDifference', [JSON.stringify(difference)]));
 
+    // If there are no differences, mark as not having OmniStudio content to exclude from report
+    // Only exclude if status is 'Ready for migration' or 'Successfully migrated' (no warnings/errors)
+    if (
+      difference.length === 0 &&
+      (experienceSiteAssessmentInfo.status === 'Ready for migration' ||
```
sf-kishore-kurri, line 184: when diff == 0 means, file is ready for migration know? why we need explicit status check here @sf-aastha-paruthi ?

sf-aastha-paruthi: This change is to exclude the record from the report. Only if the file has no changes and is of the success category we are excluding from the report.

If the file has Needs Manual Intervention status then it should still be shown on the report.

PR #457, opened 2025-12-11
@ W-20470731 Execution should stop during assessment if metadata config records are present in an org where OmniStudio Metadata is disabled.

sf-aastha-paruthi:
### What does this PR do?
image


src/migration/dataraptor.ts
```diff
@@ -296,9 +296,6 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
   public async assess(): Promise<DataRaptorAssessmentInfo[]> {
     try {
       if (isStandardDataModelWithMetadataAPIEnabled()) {
-        Logger.log(
-          this.messages.getMessage('skippingAssessmentForStandardOrgWithMetadataAPIEnabled', [Constants.DataMapper])
```
sf-kishore-kurri: this change exists n other PR also

src/commands/omnistudio/migration/assess.ts
```diff
@@ -100,10 +101,17 @@ export default class Assess extends OmniStudioBaseCommand {
     const namespace = orgs.packageDetails.namespace;
     let projectPath = '';
     const preMigrate: PreMigrate = new PreMigrate(namespace, conn, this.logger, messages, this.ux);
+    const userActionMessages: string[] = [];
 
     // Handle all versions prerequisite for standard data model
     if (isStandardDataModel()) {
```
sf-kishore-kurri: why we need double check here? The new API is already validating std data model so we can clean up

messages/assess.json
```diff
@@ -225,5 +225,6 @@
   "omniStudioAllVersionsProcessingConsentGiven": "You've consented to assess all versions of the Omnistudio components.",
   "preparingStorageForMetadataEnabledOrg": "Preparing storage for %s org on Standard Data Model with Metadata API enabled",
   "processingNotRequired": "Assessment not required for Standard Data Model Orgs with Metadata API enabled",
-  "skippingTruncation": "Skipping Trunction as standard data model org"
+  "skippingTruncation": "Skipping Trunction as standard data model org",
+  "cleanupMetadataTablesRequired": "OmniStudio metadata tables contain records that need to be cleaned up before migration. Please run the cleanup process manually to clear the metadata tables.\n<a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Cleanup Metadata Tables</a>"
```
sf-kishore-kurri, line 231: Is the message reviewed with CX?

src/commands/omnistudio/migration/assess.ts
sf-aastha-paruthi: If metadata api is off and tables have data then only we need this validation.

sf-aastha-paruthi:
Add test cases

messages/assess.json
sf-aastha-paruthi: Added to CX review.

src/utils/dataModelService.ts
```diff
@@ -9,73 +8,62 @@ export class DataModelService {
...
 }
 
-export function isFoundationPackage(): boolean {
-  if (cachedIsFoundationPackage === null) {
-    cachedIsFoundationPackage = getDataModelService().isFoundationPackage();
-  }
-  return cachedIsFoundationPackage;
+// Convenience function to check if data model is custom
+export function isCustomDataModel(): boolean {
+  return getDataModelService().checkIfIsCustomDataModel();
```
sf-kishore-kurri, line 54: getDataModelService can return error, can we handle the error appropriate return value

sf-kishore-kurri (commented):
@sf-aastha-paruthi when data model service fails, what is the expected behaviour? Does user will get feedback via error handling with action item?

Remaining changes looks good to me

sf-kishore-kurri (approved):
Approved with comment

sf-aastha-paruthi:
> @sf-aastha-paruthi when data model service fails, what is the expected behaviour? Does user will get feedback via error handling with action item?
> 
> Remaining changes looks good to me

image
The data Model service will fail in the scenario if it is not initialized via the function initialize data model service. Hardcoded the class to throw error and this is how it will appear as shown in the attached image.
This will happen if there is a code bug and initializeDataModelService is not called and direclty the other exported functions are accessed.

PR #458, opened 2025-12-17
@W-20470731 Contd Execution should stop during assessment if metadata config records are present in an org where OmniStudio Metadata is disabled and API response to handle obejct and array.

sf-aastha-paruthi:
### What does this PR do?
Fixing the function signature which was resulting in build failures.
The API calls to handle the return of both object or array data type.

### DIFF
--- a/src/utils/OmnistudioSettingsPrefManager.ts
+++ b/src/utils/OmnistudioSettingsPrefManager.ts
@@ -11,7 +11,7 @@ export class OmnistudioSettingsPrefManager {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any
       const result = (await (this.connection.metadata.read as any)('OmniStudioSettings', ['OmniStudio'])) as unknown;
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const metadata = result as MetadataInfo;
+      const metadata = (Array.isArray(result) ? result[0] : result) as MetadataInfo;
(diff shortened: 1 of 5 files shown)


src/utils/OmnistudioSettingsPrefManager.ts
```diff
@@ -48,7 +48,7 @@ export class OmnistudioSettingsPrefManager {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any
       const result = (await (this.connection.metadata.read as any)('OmniStudioSettings', ['OmniStudio'])) as unknown;
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      const metadata = result as MetadataInfo;
+      const metadata = (Array.isArray(result) ? result[0] : result) as MetadataInfo;
```
sf-kishore-kurri, line 51: @sf-aastha-paruthi Is the change effect of version upgrade? We didn't change anything on response at server end

sf-aastha-paruthi: This is needed because of version upgrade. The jsforce has some update and now the response is coming in array format.
This change ins backward compatible.

Its the same issue which I discussed that Metadata API is enabled but coming as disabled.

sf-kishore-kurri: @sf-aastha-paruthi Do we have any test cases to validate the response?

sf-kishore-kurri (approved):
Approving the changes, but we need to find away to come up with integration tests t find out the API contract changes with tests

PR #476, opened 2026-03-25
Remove automated cleanup of config tables during migration

rajender-kumar-salesforce:
Assess:
Screenshot 2026-03-26 at 5 24 09 PM

Migrate:
Screenshot 2026-03-26 at 5 25 04 PM


sf-sayeeduddin-syed (approved):

src/migration/premigrate.ts
```diff
@@ -200,14 +200,11 @@ export class PreMigrate extends BaseMigrationTool {
...
       return true;
     }
-
-    const consent = await this.getMetadataCleanupConsent();
-
-    if (!consent) {
-      Logger.error(this.messages.getMessage('metadataCleanupConsentNotGiven'));
-      return false;
-    }
-    return await omniStudioMetadataCleanupService.cleanupOmniStudioMetadataTables();
```
sf-aastha-paruthi, line 210: nit, function can be removed since no longer needed

sf-aastha-paruthi (approved):

src/migration/premigrate.ts
rajender-kumar-salesforce: Just kept it for future use case if we want to bring automated cleanup.

PR #479, opened 2026-04-27
@W-22210612  Custom data model - Omniscript Web URL in Flexcard handling

sf-aastha-paruthi:
### What does this PR do?
When we have the custom data model and we embedd and OS link in flexcard, it points to custom data model.
When we migrate we move to standard data model. So the link should also point to the standard data model Omniscript. 
The changes add this functionality.
image

image


sf-aastha-paruthi:
invoke_review

messages/assess.json
```diff
@@ -97,6 +97,7 @@
   "cardNameChangeMessage": "The Flexcard name will be changed from %s to %s to adhere to the API naming standards.",
   "authordNameChangeMessage": "The Flexcard author name will be changed from %s to %s to adhere to the naming rules",
   "cardLWCNameChangeMessage": "The Flexcard generated LWC name will be changed from %s to %s to align with Flexcard name change.",
+  "webPageOmniScriptNavigationDetected": "Detected OmniScript navigation in Web Page action URL. URL: %s.\nIt will get replaced with: %s",
```
mdmehran-qureshi: CX approved?

messages/migrate.json
```diff
@@ -201,6 +201,10 @@
   "startingComponentPreProcessing": "Pre-processing components for name mapping",
   "completeComponentMappingMessage": "Registered name mappings for %s components",
   "componentMappingNotFound": "No registry mapping found for %s component: %s, using fallback cleaning",
+  "flexCardOmniScriptNavigateUrlWillBeUpdated": "OmniScript Navigate URL will be updated to standard URL format at %s.",
```
mdmehran-qureshi: Same

messages/assess.json
sf-kishore-kurri: @sf-aastha-paruthi Are these messages are CX approved?

sf-kishore-kurri: As we are keeping full URL seems too much for me, lets review with CX on it

src/migration/flexcard.ts
```diff
@@ -700,7 +700,11 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
     return childs;
   }
 
-  private checkComponentForDependencies(component: any, flexCardAssessmentInfo: FlexCardAssessmentInfo): void {
+  private checkComponentForDependencies(
+    component: any,
+    flexCardAssessmentInfo: FlexCardAssessmentInfo,
```
sf-kishore-kurri, line 706: Instead of adding new param input, can we keep omniScriptNavigateUrlWarningKeys on FlexCardAssessmentInfo?

src/migration/flexcard.ts
```diff
@@ -814,12 +830,49 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
       }
     }
   }
 
+  private addOmniScriptNavigateUrlAssessmentWarning(
+    stateAction: any,
+    flexCardAssessmentInfo: FlexCardAssessmentInfo,
+    warningKeys: Set<string>
+  ): void {
+    if (this.IS_STANDARD_DATA_MODEL) {
```
sf-kishore-kurri, line 846: @sf-aastha-paruthi For core designer, we always give relative path only?

src/migration/flexcard.ts
```diff
@@ -1623,11 +1700,123 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+      return;
+    }
+    const { targetName, updatedUrl } = rewrite;
+
+    stateAction[Constants.WebPageTargetType].targetName = updatedUrl;
+    urlUpdateSummaries.add(this.formatUrlRewriteSummary(targetName, updatedUrl));
+  }
+
+  private formatUrlRewriteSummary(originalUrl: string, updatedUrl: string): string {
+    return `${this.toReadableUrl(originalUrl)} -> ${this.toReadableUrl(updatedUrl)}`;
```
sf-kishore-kurri: rename Url -> URL acronym

src/migration/flexcard.ts
```diff
@@ -1623,11 +1700,123 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+      Constants.OmniScriptUniversalPageToken,
+      Constants.OmniScriptTypeParam,
+      Constants.OmniScriptSubTypeParam,
+      Constants.OmniScriptLangParam,
+    ].every((token) => url.includes(token));
+  }
+
+  private convertToStandardOmniScriptUrl(inputUrl: string): string {
+    // Placeholder base lets URL parse both absolute and relative input URLs uniformly.
+    const parsedUrl = new URL(inputUrl, 'https://placeholder.local');
```
sf-kishore-kurri: what is this placeholder for? Sandbox or Developer sandbox URLs will not have .local pattern right?

sf-kishore-kurri (commented):
Added comments @sf-aastha-paruthi

messages/migrate.json
```diff
@@ -201,6 +201,10 @@
...
+  "webPageOmniScriptNavigationDetected": "Detected OmniScript navigation in Web Page action URL. URL: %s.\nIt will get replaced with: %s",
```
mdmehran-qureshi: Issue: Same key and message in both files is confusing - are they intentionally identical or accidental duplication?

test/migration/flexcard-custom-datamodel.test.ts
```diff
@@ -0,0 +1,267 @@
+/* eslint-disable @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-return, camelcase, comma-dangle */
```
mdmehran-qureshi, line 1: Missing test scenarios:

✓ Absolute URLs (covered)
✓ Relative URLs (covered via Fragment test)
✗ URLs with duplicate params (query + fragment have same param)
✗ URL with layout param renamed to omniscript__theme
✗ Malformed/invalid URLs
✗ URLs with Unicode characters
✗ Very long URLs
✗ Multiple URL actions in one card

src/migration/flexcard.ts
mdmehran-qureshi: Issue: Using a placeholder base URL allows relative URLs to be parsed, but this could mask issues:

What if inputUrl is javascript:alert(1) or similar protocol handler?
The URL constructor will reject invalid URLs, but edge cases should be tested
No validation that the parsed URL has expected components (path must contain /apex/)

src/migration/flexcard.ts
```diff
@@ -814,12 +830,49 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
 
     // Check child components recursively
     if (component.children && Array.isArray(component.children)) {
```
mdmehran-qureshi, line 833: No maximum depth limit - could cause stack overflow on deeply nested/circular structures

src/migration/flexcard.ts
mdmehran-qureshi: Issue:

If inputUrl contains invalid URL characters, this will throw
No try-catch wrapping this logic
Calling code (getOmniScriptUrlRewrite()) doesn't handle exceptions

messages/assess.json
sf-aastha-paruthi: I have added then to the doc.
https://docs.google.com/document/d/1Qo6yzNxWYMgxWPD1yjMpUuX-wDzgdhQGo_DfPu_yFno/edit?tab=t.0
Will update once get CX inputs.

src/migration/flexcard.ts
sf-aastha-paruthi: We are keeping a param to keep track of the urls and ensuring that there are no duplicates. Once the complete set is prepared it is appended to the FlexcardAssessmentInfo.

src/migration/flexcard.ts
sf-aastha-paruthi: URLs can be both relative and abosolute in both package and core.
We are replacing with relative URLs in standard world.

sf-aastha-paruthi: This Placeholder base lets URL parse both absolute and relative input URLs uniformly.

messages/assess.json
sf-aastha-paruthi: These are added then to the doc.
https://docs.google.com/document/d/1Qo6yzNxWYMgxWPD1yjMpUuX-wDzgdhQGo_DfPu_yFno/edit?tab=t.0
Will update once get CX inputs.

messages/migrate.json
sf-aastha-paruthi: Same

messages/migrate.json
sf-aastha-paruthi: Yes, the key needs to be added in both the places as both access and migrate use different file to pick these keys.

src/migration/flexcard.ts
sf-aastha-paruthi: If we are on the Standard data model we do not need this change. Only when we have the custom omniscript URLs which get shifted to standard we need to take care of it.
When we are updating we are giving relative path only as it is supported.

src/migration/flexcard.ts
```diff
@@ -1623,11 +1700,123 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+    }
+
+    const newSearchParams = new URLSearchParams();
+    for (const [key, value] of mergedEntries) {
+      const newKey = paramRenames[key] ?? key;
+      const newValue = paramsToClean.has(key) ? this.cleanName(value) : value;
+      newSearchParams.append(newKey, newValue);
+    }
+
+    return `${Constants.OmniScriptStandardPagePath}?${newSearchParams.toString()}`;
```
sf-aastha-paruthi, line 1967: Can you please tell which condition?

src/migration/flexcard.ts
```diff
@@ -1623,11 +1701,196 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+    }
+    const tokens = trimmed.split('/').map((t) => {
+      try {
+        return decodeURIComponent(t);
+      } catch {
+        return t;
+      }
+    });
+    const pairs: Array<[string, string]> = [];
+    for (let i = 0; i + 1 < tokens.length; i += 2) {
```
shaurabh-tiwari-git, line 2001: What if the tokens length is not even?

src/migration/flexcard.ts
```diff
@@ -1809,7 +1877,13 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
     if (!trimmed) {
       return [];
     }
-    const tokens = trimmed.split('/').map((t) => decodeURIComponent(t));
+    const tokens = trimmed.split('/').map((t) => {
```
sf-kishore-kurri, line 1992: @sf-aastha-paruthi As discussed over call, when URL has invalid values instead of skipping and using the same, lets identify and mark them into either warning or manual bucket so that customer is aware of the issue and take necessary actions

src/migration/flexcard.ts
```diff
@@ -1705,64 +1706,126 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+   * `javascript:alert(1)//apex/...` whose parsed pathname does not begin
+   * with `/` (it begins with the opaque body of the scheme).
+   *
+   * Two legitimate shapes are accepted:
+   *   1. Path + query:    /apex/<ns>__OmniScriptUniversalPage?OmniScriptType=...&OmniScriptSubType=...&OmniScriptLang=...
+   *   2. Path + fragment: /apex/<ns>__OmniScriptUniversalPage?...#/OmniScriptType/.../OmniScriptSubType/.../OmniScriptLang/...
+   */
+  private isValidOmniScriptNavigationURL(parsedURL: URL): boolean {
+    if (!parsedURL.pathname.startsWith(`${Constants.OmniScriptUniversalPagePath}/`)) {
+      return false;
```
sf-kishore-kurri, line 1921: If we can maintain the invalidURLs list and valid against common logic using regex would be easy giong forward, but please check whether this is feasible

mdmehran-qureshi (approved):
resolve existing comments first.

src/migration/flexcard.ts
sf-aastha-paruthi: Done

src/migration/flexcard.ts
sf-aastha-paruthi: Done

sf-aastha-paruthi: Done

src/migration/flexcard.ts
sf-aastha-paruthi: It's handled. The loop guard is i + 1 < tokens.length (not i < tokens.length), so when the token count is odd the trailing orphan token is skipped rather than producing a [key, undefined] pair. This matches the documented behavior in the JSDoc ("trailing orphan token (which is dropped)"), since the fragment format is strictly #/Key1/Value1/Key2/Value2/... and a lone trailing key has no meaningful value to pair with.

shaurabh-tiwari-git: got it

shaurabh-tiwari-git (approved):

sf-kishore-kurri (approved):
LGTM

PR #481, opened 2026-05-14
@W-19747584 - Custom CSS Scanning and Assessment for Flexcard and Omniscript

shaurabh-tiwari-git:
Problem - During the migration we are changing the patterns at several places where the occurences of namespaces are getting changed to c (For example: vlocity_ins/function -> c/function etc). In doing so the html structure will change and if any customer has implemented his custom css which has the namespace dependency might break.

Solution - For this scenario we are scanning through the custom css files used in Omniscripts and Flexcards, and showing a warning to the user during assessment if there are any occurrences of the namespace in the css file. There might be other patterns possible which wee are not certain of but this PR will cover the namespace scenario.

### DIFF
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -188,6 +188,7 @@
   "nameMappingUndefined": "Undefined name mapping found",
   "experienceSiteException": "We’ve encountered an exception while processing Experience Cloud sites.",
   "reservedKeysFoundInPropertySet": "The Integration Procedure output uses these reserved keys in the transformation property fields: %s",
+  "customCssStylesheetNamespaceWarning": "Custom CSS stylesheet '%s' has namespace references, styles may break after migration.",
   "dataMapperMigrationFailed": "DataMapper migration failed for: %s",
   "dataMapperNameStartsWithNumber": "DataMapper name '%s' starts with a number which causes migration issues. Proposed new name: %s",
   "integrationProcedureTypeEmptyAfterCleaning": "Integration Procedure Type '%s' becomes empty after name cleaning. Please provide a valid Type value.",
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -189,6 +189,7 @@
   "experienceSiteException": "We’ve encountered an exception while processing Experience Cloud sites.",
   "reservedKeysFoundInPropertySet": "The Integration Procedure output uses these reserved keys in the transformation property fields: %s",
   "customCssStylesheetNamespaceWarning": "Custom CSS stylesheet '%s' has namespace references, styles may break after migration.",
+  "customCssInlineNamespaceWarning": "Custom CSS has namespace references, styles may break after migration.",
   "dataMapperMigrationFailed": "DataMapper migration failed for: %s",
   "dataMapperNameStartsWithNumber": "DataMapper name '%s' starts with a number which causes migration issues. Proposed new name: %s",
   "integrationProcedureTypeEmptyAfterCleaning": "Integration Procedure Type '%s' becomes empty after name cleaning. Please provide a valid Type value.",
--- a/src/migration/CustomCssRegistry.ts
+++ b/src/migration/CustomCssRegistry.ts
@@ -1,0 +1,217 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* eslint-disable */
+import { Connection, Messages } from '@salesforce/core';
+import { QueryTools } from '../utils/query';
+import { Logger } from '../utils/logger';
+
+/**
+ * Outcome of scanning a single Static Resource for namespace references.
+ *  - `notFound`    — no record with this Name exists in the org
+ *  - `unsupported` — body was unreadable or non-text & non-zip
+ *  - `clean`       — fetched, scanned, no namespace match
+ *  - `dirty`       — fetched, scanned, namespace match found
+ */
+export type CustomCssScanVerdict = 'notFound' | 'unsupported' | 'clean' | 'dirty';
+
+/**
+ * Result of scanning the four-key `stylesheet` object embedded in an OmniScript's
+ * (or FlexCard's) PropertySetConfig. Returns the names of stylesheets whose
+ * StaticResource bodies still reference the org's managed-package namespace.
+ *
+ * Callers use `dirtyStylesheets` to push warnings; the array preserves insertion
+ * order so messages reflect the variant order checked (lightning, newport, …).
+ */
+export interface CustomCssStylesheetScanResult {
+  dirtyStylesheets: string[];
+}
+
+/**
+ * Centralised, process-wide cache & scanner for Static Resources referenced as
+ * "Custom Lightning / Newport Stylesheet File Name" in OmniScript / FlexCard
+ * PropertySetConfig.
+ *
+ * Why a registry (vs a per-tool field):
+ *  - The same StaticResource is frequently shared across many OmniScripts and
+ *    (in future) FlexCards. We don't want to re-query / re-download it per tool.
+ *  - OmniScript and Integration Procedure are two separate
+ *    `OmniScriptMigrationTool` instances in the assess flow — without the
+ *    registry, their caches would be siloed.
+ *  - Future component types (FlexCard, etc.) can plug in by calling
+ *    `scanResource()` directly without re-implementing the fetch/zip/scan logic.
+ *
+ * Lifetime: a single assessment run. Tools call `init(connection, namespace,
+ * messages)` once at construction time; subsequent inits are idempotent (the
+ * first one wins, and a `reset()` is exposed for tests / re-entrant runs).
+ */
+export class CustomCssRegistry {
+  private static instance: CustomCssRegistry;
+
+  private cache: Map<string, CustomCssScanVerdict> = new Map();
+  private connection?: Connection;
+  private namespace?: string;
+  private messages?: Messages<string>;
+
+  /** Variant keys present on `propertySetConfig.stylesheet` for OmniScript. */
+  private static readonly OMNISCRIPT_STYLESHEET_VARIANTS: ReadonlyArray<
+    'lightning' | 'newport' | 'lightningRtl' | 'newportRtl'
+  > = ['lightning', 'newport', 'lightningRtl', 'newportRtl'];
+
+  public static getInstance(): CustomCssRegistry {
+    if (!CustomCssRegistry.instance) {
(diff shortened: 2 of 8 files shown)


messages/assess.json
```diff
@@ -188,6 +188,8 @@
   "nameMappingUndefined": "Undefined name mapping found",
   "experienceSiteException": "We’ve encountered an exception while processing Experience Cloud sites.",
   "reservedKeysFoundInPropertySet": "The Integration Procedure output uses these reserved keys in the transformation property fields: %s",
+  "customCssStylesheetNamespaceWarning": "Custom CSS stylesheet '%s' has namespace references, styles may break after migration.",
```
sf-kishore-kurri, line 191: CSS is acronym

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* eslint-disable */
+import { Connection, Messages } from '@salesforce/core';
+import { QueryTools } from '../utils/query';
+import { Logger } from '../utils/logger';
+
```
sf-kishore-kurri, line 12: @shaurabh-tiwari-git Are we migrating the custom css which are not compatible with core runtime or only assess?

src/migration/flexcard.ts
```diff
@@ -317,6 +324,92 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+   *     scans are cached by name in {@link CustomCssRegistry}, so a stylesheet
+   *     shared between OmniScripts and FlexCards is only fetched once per run.
+   *
+   *  2. **Inline CSS** — `Styles__c.customStyles` stores raw CSS text. We
+   *     substring-check the namespace directly against that string. No
+   *     SOQL/REST traffic and no caching (the CSS is per-FlexCard and not
+   *     shared across records).
+   *
+   * Both checks emit a warning into `flexCardAssessmentInfo.warnings` and bump
+   * `migrationStatus` to 'Warnings' (using `getUpdatedAssessmentStatus` so a
```
sf-kishore-kurri, line 354: @shaurabh-tiwari-git We should run your fix against the Prudential sandbox once to validate the fix

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Only assess. We discussed it earlier as well that there might be several patterns and we will not be able to migrate them.

src/migration/flexcard.ts
shaurabh-tiwari-git: Can you share the org details with me?

messages/assess.json
shaurabh-tiwari-git: I've added these messages for CX review document and will update once the approved message comes.

sf-pallavi-das (approved):
LGTM

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+        for (const file of entries) {
+          if (file.dir) continue;
+          if (!/\.css$/i.test(file.name)) continue;
+          const cssText: string = await file.async('string');
+          if (containsNamespace(cssText)) {
+            dirty = true;
+            break;
+          }
+        }
+        const verdict: CustomCssScanVerdict = dirty ? 'dirty' : 'clean';
```
sf-pallavi-das: What does dirty mean here? Does it mean custom stylesheet? If yes, it's looks misleading

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const sheet = stylesheet as Record<string, unknown>;
+    for (const variant of CustomCssRegistry.OMNISCRIPT_STYLESHEET_VARIANTS) {
+      const raw = sheet[variant];
+      const resourceName = (raw == null ? '' : String(raw)).trim();
+      if (!resourceName) continue;
+
+      const verdict = await this.scanResource(resourceName);
+      if (verdict === 'dirty') {
+        result.dirtyStylesheets.push(resourceName);
```
rajender-kumar-salesforce: can we dedupe dirtyStylesheets to avoid duplicate warnings Use set.

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
+ * For full license text, see LICENSE.txt file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/* eslint-disable */
```
rajender-kumar-salesforce: Disabling eslint for entire file is not recomended.

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+        return verdict;
+      }
+
+      // Zip archive — scan every .css entry.
+      if (contentType === 'application/zip' || contentType === 'application/x-zip-compressed') {
+        const raw = await this.connection.request<any>({ method: 'GET', url });
+        const buf: Buffer = Buffer.isBuffer(raw) ? raw : Buffer.from(raw, 'binary');
+        // Lazy-require so jszip only loads when a zipped resource is actually encountered.
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const JSZip = require('jszip');
```
rajender-kumar-salesforce: Any Idea why this inline required instead of import JSZip from 'jszip' ?

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    // 1. Look up the record. StaticResource is a standard sObject — no namespace prefix.
+    const filters = new Map<string, any>([['Name', resourceName]]);
+    const rows = (await QueryTools.query(
+      this.connection,
+      'StaticResource',
+      ['Id', 'Name', 'ContentType', 'BodyLength'],
+      filters
+    )) as any[];
```
rajender-kumar-salesforce: Use interface.

src/migration/flexcard.ts
```diff
@@ -317,6 +324,92 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+   * stricter status set elsewhere is preserved).
+   */
+  private async collectStylesheetNamespaceDependencies(
+    flexCard: AnyJson,
+    flexCardAssessmentInfo: FlexCardAssessmentInfo
+  ): Promise<void> {
+    const registry = CustomCssRegistry.getInstance();
```
rajender-kumar-salesforce, line 361: Is there action item or help link for customer which tells what to do in this scenario ?

messages/assess.json
```diff
@@ -188,6 +188,8 @@
...
+  "customCssInlineNamespaceWarning": "Custom CSS has namespace references, styles may break after migration.",
```
rajender-kumar-salesforce, line 192: We should include which namespace ?

src/migration/flexcard.ts
```diff
@@ -317,6 +324,92 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
...
+        definition = JSON.parse(definitionRaw);
+      } catch (ex) {
+        Logger.error(`Failed to parse Definition__c for stylesheet scan: ${flexCard['Name']}`);
+        definition = null;
+      }
+      const resourceName = (definition?.customStyleSheet || '').toString().trim();
+      if (resourceName) {
+        const verdict = await registry.scanResource(resourceName);
+        if (verdict === 'dirty') {
+          const message = registry.buildNamespaceWarning(resourceName);
```
rajender-kumar-salesforce, line 380: Should we keep it as warnings or needs manual intervention ?

src/migration/flexcard.ts
shaurabh-tiwari-git: CTA links for corresponding document will be added, that will be a different user story and will be taken in upcoming sprint.

messages/assess.json
shaurabh-tiwari-git: In the beginning of the assessment we either select the namespace or it gets automatically selected (for single namespace), we are using that namespace.

src/migration/CustomCssRegistry.ts
```diff
@@ -0,0 +1,240 @@
...
+        // Lazy-require so jszip only loads when a zipped resource is actually encountered.
+        // eslint-disable-next-line @typescript-eslint/no-var-requires
+        const JSZip = require('jszip');
+        const zip = await JSZip.loadAsync(buf);
+        let dirty = false;
+        const entries: any[] = Object.values(zip.files);
+        for (const file of entries) {
+          if (file.dir) continue;
+          if (!/\.css$/i.test(file.name)) continue;
+          const cssText: string = await file.async('string');
```
rajender-kumar-salesforce: We should log file name which is dirty or possibly we can add in warning message itself ? Zip may have 100 files and customer need to know which file has issue.

src/migration/flexcard.ts
shaurabh-tiwari-git: Since this is only a styling issue and not a functionality break, keeping this as warnings will be better according to me.

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: We are adding the warning message against the os/fc which has the css with namespace and also we are including the name of the static resource css stylesheet. There is one case in flexcard where we can directly write custom css in the builder, there we are mentioning about the custom css.

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Addressed

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Addresssed

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Addressed

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Addressed

rajender-kumar-salesforce (approved):

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: earlier it was here so that it will be required only when there is a static resource in the fc/os, but that'll handled by node anyways in case of import and also it'll be required once not per call. So changed it to top level import.

src/migration/CustomCssRegistry.ts
shaurabh-tiwari-git: Addressed

PR #487, opened 2026-05-20
@W-22403162: Auto-enable PreHook/PostHook on IPs migrated from managed package

adamespii:
## Summary
When migrating Integration Procedures from managed-package runtime to Core standard runtime, customers silently lose their custom pre/post hook logic because the new PreHook/PostHook booleans default to false. This PR patches the migration tool to detect registered CpqAppHandlerHook implementations in the source org and auto-enable both flags on every migrated Remote Action step, preserving managed-package parity without manual intervention.

## 🔍 Root Cause Analysis

**Problem:** Customers with a CpqAppHandlerHook registered in CustomClassImplementation__c lose hook execution after IP migration because the Core standard runtime requires explicit opt-in via PreHook/PostHook boolean flags on each Remote Action step's remoteOptions, whereas the managed-package runtime fired hooks automatically whenever a hook was registered.

**Primary cause:** The migration script did not inspect the source org's hook registrations or set the corresponding Core-side flags during element migration.

**Secondary cause:** The Core-side PreHook/PostHook fields intentionally default to false for greenfield IPs (correct for new development, incorrect for migrated IPs that relied on implicit hook behavior).

**Severity:** High - Migrated customers silently lose business-critical hook logic (CPQ pricing, validation, enrichment) with no runtime error, only incorrect behavior.

## 🔧 Solution

- Added `checkCpqAppHandlerHook()` method that queries `{namespace}__CustomClassImplementation__c` for a `CpqAppHandlerHook` record in the source org, cached once per migration run
- Added `case 'Remote Action':` to the `mapElementData` switch that sets `remoteOptions.PreHook = true` and `remoteOptions.PostHook = true` when the hook is registered
- Query failure (permission issues, missing object) gracefully defaults to `false` with a warning log — migration continues without setting hooks rather than aborting
- The hook check runs once in `migrate()` before the element processing loop (not per-element), stored in `this.hasCpqAppHandlerHook`
- The new case mutates the already-parsed `propertySet` object in place; the existing `JSON.stringify` after the switch handles serialization

**Files changed:**
- `src/migration/omniscript.ts`
- `test/migration/omniscript.test.ts` (new)

## 🧪 Testing

Added 5 Mocha test scenarios covering:
1. Hook registered → PreHook=true, PostHook=true set on Remote Action elements
2. No hook registered → remoteOptions unchanged (no keys added)
3. Flags already true + hook registered → values preserved (idempotent)
4. Non-Remote-Action element type → not affected by this change
5. CustomClassImplementation__c query failure → defaults to false, logs warning, migration continues

Static analysis: ESLint PASSED (pre-existing warning in unrelated file only), TypeScript `tsc --noEmit` PASSED

## Impact Assessment

- **Breaking changes:** None — existing migration behavior for all non-Remote-Action element types is unchanged
- **Performance impact:** One additional SOQL query per migration run (cached result), negligible
- **Related WIs:** W-22403162 (this story), TD-0271459 (parent initiative), W-22202779–W-22202839 (Core-side PreHook/PostHook implementation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [1.6.1](https://github.com/salesforcecli/plugin-omnistudio-migration-tool/compare/1.6.0...1.6.1) (2025-09-12)
+
+
+
 # [1.6.0](https://github.com/salesforcecli/plugin-omnistudio-migration-tool/compare/1.5.0...1.6.0) (2024-09-12)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+# [1.7.0](https://github.com/salesforcecli/plugin-omnistudio-migration-tool/compare/1.6.1...1.7.0) (2025-11-07)
+
+
+### Features
+
+* @W-19700286: [OMT] [Migration] Implemention of Guardrail and validation ([55fd750](https://github.com/salesforcecli/plugin-omnistudio-migration-tool/commit/55fd75039d92e42198580a577279c78122207400))
+
+
+
 ## [1.6.1](https://github.com/salesforcecli/plugin-omnistudio-migration-tool/compare/1.6.0...1.6.1) (2025-09-12)
 
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 # OmniStudio Migration Tool
 
-### Before You Begin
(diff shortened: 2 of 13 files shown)


adamespii:
Moved to https://github.com/salesforcecli/plugin-omnistudio-migration-tool/pull/490

PR #488, opened 2026-05-21
@W-19747619 - Detect corrupted parent-child level hierarchy in OmniScript elements

mdmehran-qureshi:
## Summary
- Adds pre-migration detection for OmniScripts with corrupted parent-child element hierarchy where parent and child elements are persisted at the same `Level__c` value (both at level 0)
- During **assessment**, corrupted OmniScripts are flagged as "Needs manual intervention" with a descriptive warning listing all affected element names
- During **migration**, corrupted OmniScripts are **skipped** entirely to prevent silent data loss

## Context
- Root cause: OmniScripts with Action Block elements in the managed package data model have data corruption where parent and child elements are persisted at the same level (level 0)
- Impact: Elements are lost/missing after migration from managed package runtime to OmniStudio standard runtime
- Workaround: Creating a new version of the OmniScript via the UI fixes the corrupted hierarchy (confirmed via manual testing)

## Test plan
- [x] Unit tests added (`test/migration/omniscript-corrupted-parent-child.test.ts`) - 6 test cases covering assessment and migration modes
- [x] All 622 existing tests pass
- [x] Manual testing in sandbox org (`cmtlatest`) confirmed:
  - Corrupted OmniScript correctly flagged with error message listing affected elements
  - Valid OmniScript passes assessment with `migrationSuccess: true`
  - UI "Create Version" confirmed to fix the `Level__c` hierarchy (child elements move from level 0 to level 1)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -238,5 +238,5 @@
   "apexRemoteDatasourceNamespaceWarning": "ApexRemote datasource: remoteClass will be updated from \"%s\" to \"%s\"",
   "apexClassNotFound": "Apex class \"%s\" referenced in \"%s\" does not exist in the org",
   "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>",
-  "corruptedParentChildLevel": "OmniScript has elements with corrupted parent-child hierarchy. The following elements have their parent and child persisted at the same level, which will cause data loss after migration: %s. Create a new version of this OmniScript to fix the issue before migrating."
+  "corruptedParentChildLevel": "%s has elements with corrupted parent-child hierarchy. The following elements have their parent and child persisted at the same level, which will cause data loss after migration: %s. Create a new version of this %s to fix the issue before migrating."
 }
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -237,5 +237,6 @@
   "remoteActionNamespaceWarning": "Remote Action \"%s\": remoteClass will be updated from \"%s\" to \"%s\"",
   "apexRemoteDatasourceNamespaceWarning": "ApexRemote datasource: remoteClass will be updated from \"%s\" to \"%s\"",
   "apexClassNotFound": "Apex class \"%s\" referenced in \"%s\" does not exist in the org",
-  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>"
+  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>",
+  "corruptedParentChildLevel": "OmniScript has elements with corrupted parent-child hierarchy. The following elements have their parent and child persisted at the same level, which will cause data loss after migration: %s. Create a new version of this OmniScript to fix the issue before migrating."
 }
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -707,7 +707,9 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     // Add warning for corrupted parent-child level hierarchy (elements at same level as their parent)
     if (corruptedParentChildElements.size > 0) {
       const corruptedNamesList = Array.from(corruptedParentChildElements).join(', ');
-      warnings.push(this.messages.getMessage('corruptedParentChildLevel', [corruptedNamesList]));
+      warnings.push(
+        this.messages.getMessage('corruptedParentChildLevel', [omniProcessType, corruptedNamesList, omniProcessType])
+      );
       assessmentStatus = 'Needs manual intervention';
     }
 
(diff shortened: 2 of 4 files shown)


src/migration/omniscript.ts
```diff
@@ -433,6 +436,24 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       }
     }
 
+    // Build a map of element IDs to their levels for parent-child validation
+    const elementLevelMap = new Map<string, number>();
+    for (const elem of elements) {
+      elementLevelMap.set(elem['Id'], elem[this.getElementFieldKey('Level__c')]);
+    }
+
+    // Detect elements that have a ParentElementId but share the same level as their parent
```
shaurabh-tiwari-git: Can we use the detectCorruptedParentChildElements(elements: AnyJson[]) function to avoid code duplication here?

src/migration/omniscript.ts
```diff
@@ -433,6 +436,24 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+    for (const elem of elements) {
+      const parentId = elem[this.getElementFieldKey('ParentElementId__c')];
+      if (parentId && elementLevelMap.has(parentId)) {
+        const childLevel = elem[this.getElementFieldKey('Level__c')];
+        const parentLevel = elementLevelMap.get(parentId);
+        if (childLevel === parentLevel) {
```
shaurabh-tiwari-git: is null or undefined handling required here?

src/migration/omniscript.ts
```diff
@@ -701,6 +722,13 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       assessmentStatus = 'Needs manual intervention';
     }
 
+    // Add warning for corrupted parent-child level hierarchy (elements at same level as their parent)
+    if (corruptedParentChildElements.size > 0) {
+      const corruptedNamesList = Array.from(corruptedParentChildElements).join(', ');
+      warnings.unshift(this.messages.getMessage('corruptedParentChildLevel', [corruptedNamesList]));
```
shaurabh-tiwari-git: is there a specific reason to use unshift instead of push?

mdmehran-qureshi: No, let me change it to push.

src/migration/omniscript.ts
mdmehran-qureshi: Not required I guess, But adding for extra security.

src/migration/omniscript.ts
mdmehran-qureshi: Done

src/migration/omniscript.ts
```diff
@@ -2737,6 +2766,39 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+    const corruptedElements = new Set<string>();
+
+    // Build a map of element IDs to their levels
+    const elementLevelMap = new Map<string, number>();
+    for (const elem of elements) {
+      elementLevelMap.set(elem['Id'], elem[this.getElementFieldKey('Level__c')]);
+    }
+
+    // Check each element: if it has a parent and both are at the same level, it's corrupted
+    for (const elem of elements) {
```
sf-pallavi-das, line 2796: Does this take care of nested paremt child elements?

sf-pallavi-das (approved):
LGTM

messages/assess.json
```diff
@@ -239,5 +237,6 @@
...
   "apexClassNotFound": "Apex class \"%s\" referenced in \"%s\" does not exist in the org",
-  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>"
+  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>",
+  "corruptedParentChildLevel": "OmniScript has elements with corrupted parent-child hierarchy. The following elements have their parent and child persisted at the same level, which will cause data loss after migration: %s. Create a new version of this OmniScript to fix the issue before migrating."
```
sf-pallavi-das: Needs CX review

shaurabh-tiwari-git (approved):
LGTM

messages/assess.json
sf-kishore-kurri: Does the IP also support block component?

src/migration/omniscript.ts
sf-kishore-kurri: @mdmehran-qureshi other than block component, do we have any other elements which can work with parent/child?

messages/assess.json
mdmehran-qureshi: Yes, IPs also support Block (and other container) elements. Both OmniScript and IP use the same element structure with `ParentElementId__c` and `Level__c` fields. This detection runs for both OS and IP assessments since they share the same `assessOmniScript` code path — IPs are assessed as a subtype of OmniScript (Type = 'Integration Procedure').

src/migration/omniscript.ts
mdmehran-qureshi: Yes, the detection handles nested parent-child elements at any depth. The logic checks every element that has a `ParentElementId__c` — it compares its `Level__c` with its parent's `Level__c`. So whether the corruption is at Level 0→0, Level 1→1, or Level 2→2, it will be detected. Container elements that support children include: **Block**, **Step**, **Type Ahead Block**, **Edit Block**, and **Conditional Block**. The check is type-agnostic — it only looks at the Level/ParentElement relationship regardless of element type.

messages/assess.json
sf-kishore-kurri: then @mdmehran-qureshi message only says Omniscript which should be corrected based on type here

mdmehran-qureshi: Good catch! Fixed in commit 43a3e3a — the message now dynamically uses the component type (`OmniScript` or `Integration Procedure`) instead of hardcoding "OmniScript". The message format is now: `"%s has elements with corrupted parent-child hierarchy... Create a new version of this %s to fix the issue before migrating."`

sf-kishore-kurri (approved):
Approving the changes, As we are not migrating automatically we need to have CTA links for manual update steps. Please take care of it @mdmehran-qureshi

src/utils/constants/documentRegistry.ts
```diff
@@ -31,4 +31,5 @@ export const documentRegistry = {
     'https://help.salesforce.com/s/articleView?id=xcloud.os_standard_set_up_your_environment_for_customizing_omniscript_elements.htm&type=5',
   customLabelMigrationErrorMessage:
     'https://help.salesforce.com/s/articleView?id=xcloud.os_migrate_oma_prereq.htm&type=5',
+  corruptedParentChildLevel: 'https://help.salesforce.com/s/articleView?id=xcloud.os_version_omniscripts.htm&type=5',
```
sf-kishore-kurri, line 34: Please replace this placeholder links with CX created documentation links with manual steps to be performed by customer

PR #489, opened 2026-05-21
feat: @W-19594198 - Add CSV export to assessment and migration reports

mdmehran-qureshi:
## Summary
- Adds CSV export functionality for Custom Labels assessment and migration reports (W-19594198)
- Custom Labels HTML reports only show error/warning records (due to volume of 10K+ labels), but the CSV export includes ALL records (success + errors) for offline review
- Generates server-side CSV files (`customlabel_assessment_export.csv` / `customlabel_migration_export.csv`) at report-build time with all labels
- Adds "Export CSV" button to both assessment and migration report templates
- For other components (OS, IP, FC, DR, Apex, LWC, etc.), the export scrapes visible table rows (respects filters/search)
- Includes proper CSV escaping, HTML tag stripping, and UTF-8 BOM for Excel compatibility

## Test plan
- [x] Run `sf omnistudio migration assess` with Custom Labels and verify "Export CSV" button appears
- [x] Click "Export CSV" on Custom Labels report — verify it downloads a CSV with ALL labels (including "Ready for migration" records not shown in HTML table)
- [x] Verify CSV has correct headers: Name, Package - Id, Package - Value, Core - Id, Core - Value, Assessment Status, Summary
- [x] Open exported CSV in Excel and verify UTF-8 characters and HTML-stripped values render correctly
- [x] Run `sf omnistudio migration migrate` with Custom Labels and verify "Export CSV" downloads a CSV with all migration results
- [x] For non-Custom-Label reports (e.g., OmniScript, FlexCard), click "Export CSV" and verify it exports only visible/filtered rows from the table
- [x] Apply filters/search on a non-Custom-Label report, then export — verify only filtered rows appear in CSV

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/utils/resultsbuilder/index.ts
```diff
@@ -319,6 +319,10 @@ export class ResultsBuilder {
 
     Logger.logVerbose(messages.getMessage('generatingCustomLabelsReport', [totalLabels, totalPages, pageSize]));
 
+    // Generate CSV file with all labels for export
+    const csvFileName = 'customlabel_migration_export.csv';
```
shaurabh-tiwari-git: This string is duplicated at multiple places can me make it a global variable and use it from there?

shaurabh-tiwari-git (approved):
Please add tests for these changes, rest looks good to me.

src/utils/resultsbuilder/index.ts
mdmehran-qureshi: These are not duplicate however I moved to constant file.

src/utils/resultsbuilder/index.ts
```diff
@@ -347,6 +353,66 @@ export class ResultsBuilder {
     }
   }
 
+  /**
+   * Generates a CSV file containing all custom label migration data for export
+   */
+  // eslint-disable-next-line complexity
+  private static generateCustomLabelMigrationCsvFile(fileName: string, result: MigratedObject): void {
```
rajender-kumar-salesforce, line 360: Do we have figma design for csv export assess and migrate mode. Add expected vs real report for references.

mdmehran-qureshi: There is no figma design.

src/javascripts/reportGeneratorUtility.js
```diff
@@ -566,10 +566,136 @@ function applyDynamicStickyColumns() {
 // Run on page load
 document.addEventListener('DOMContentLoaded', applyDynamicStickyColumns);
 
+/**
+ * Export visible table data to CSV file
+ * Only exports rows that are currently visible (respects active filters and search)
+ */
+function exportTableToCSV(tableId) {
```
rajender-kumar-salesforce, line 574: Do we have any loader state meanwhile csv is getting downloaded and sucess or error message if something goes bad while exporting csv ?

mdmehran-qureshi: Are we doing that for any other reports?

mdmehran-qureshi: No, there is no loader/spinner/toast or success/error messaging anywhere in the existing report frontend. The entire reportGeneratorUtility.js uses silent catch blocks with console.error only. All report interactions (filtering, searching, CTA panel) have no loading indicators or user feedback. Additionally, the CSV download is a local file operation (pre-generated file or Blob download) — it completes instantly with no network call involved, so a loader state isn't necessary here.

rajender-kumar-salesforce: since csv report can have huge data so there should be some confirmation on code that either report download failed, succeeded or its in progress. Assessment and Migration reports are generated from terminal and generally don't need these loading state. better to check with UX for it.

mdmehran-qureshi: Good point. Added visual feedback on the Export CSV button in commit f5ab9eb:

- **In-progress**: Button shows "Exporting..." and is disabled while processing
- **Success**: Button turns green with "✓ Exported" for 2 seconds, then resets
- **Error**: Button turns red with "Export failed" for 3 seconds, then resets. Error is also logged to console.

The entire export is wrapped in a try/catch so any unexpected failure (missing table, DOM issues) is caught and surfaced to the user via the button state.

rajender-kumar-salesforce (approved):

src/javascripts/reportGeneratorUtility.js
```diff
@@ -566,10 +566,191 @@ function applyDynamicStickyColumns() {
...
+    link.setAttribute('href', url);
+    link.setAttribute('download', filename);
+    link.style.display = 'none';
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    URL.revokeObjectURL(url);
+    showExportSuccess(exportButton);
+  } catch (error) {
+    console.error('CSV export failed:', error);
```
sf-kishore-kurri, line 695: @mdmehran-qureshi Once export is failed, can we re-export or button will be changed to export failed only

mdmehran-qureshi: It will show Export after 2-3 sec

src/templates/assessmentReport.template
```diff
@@ -94,6 +94,14 @@
                 <input type="text" id="name-search-input" class="search-input" placeholder="Search Name"
                   oninput="filterAndSearchTable('report-table')" />
               </div>
+              <!-- Export to CSV button -->
+              <button class="export-csv-button" onclick="exportTableToCSV('report-table')" title="Export to CSV">
+                <svg xmlns="http://www.w3.org/2000/svg" width="14" height="14" fill="currentColor" viewBox="0 0 16 16">
+                  <path d="M.5 9.9a.5.5 0 0 1 .5.5v2.5a1 1 0 0 0 1 1h12a1 1 0 0 0 1-1v-2.5a.5.5 0 0 1 1 0v2.5a2 2 0 0 1-2 2H2a2 2 0 0 1-2-2v-2.5a.5.5 0 0 1 .5-.5"/>
```
sf-kishore-kurri, line 100: @mdmehran-qureshi Does the export button only visible for custom labels as we added in template?

mdmehran-qureshi: No, the Export CSV button is visible on **all** component reports (OmniScript, FlexCard, IP, DataMapper, Apex, LWC, etc.) — not just Custom Labels. It's placed in the shared `assessmentReport.template` and `migrationReport.template`.

The behavior differs by component:
- **Custom Labels**: Downloads a pre-generated CSV file containing ALL records (including success records not shown in HTML)
- **All other components**: Scrapes the visible table rows and exports them as CSV (respects active filters/search)

This is intentional — the WI specifically called out Custom Labels needing a separate full-data CSV because the HTML only shows errors. Other components already show all records in their HTML table, so scraping the table gives the complete picture.

PR #490, opened 2026-05-21
@W-22403162: Auto-enable PreHook/PostHook on migrated Remote Action steps

adamespii:
## Summary
When migrating Integration Procedures from managed-package runtime to Core standard runtime, customers silently lose their custom pre/post hook logic because the new PreHook/PostHook booleans default to false. This PR patches the migration tool to detect registered CpqAppHandlerHook implementations in the source org and auto-enable both flags on every migrated Remote Action step, preserving managed-package parity without manual intervention.

## 🔍 Root Cause Analysis

**Problem:** Customers with a CpqAppHandlerHook registered in CustomClassImplementation__c lose hook execution after IP migration because the Core standard runtime requires explicit opt-in via PreHook/PostHook boolean flags on each Remote Action step's remoteOptions, whereas the managed-package runtime fired hooks automatically whenever a hook was registered.

**Primary cause:** The migration script did not inspect the source org's hook registrations or set the corresponding Core-side flags during element migration.

**Secondary cause:** The Core-side PreHook/PostHook fields intentionally default to false for greenfield IPs (correct for new development, incorrect for migrated IPs that relied on implicit hook behavior).

**Severity:** High - Migrated customers silently lose business-critical hook logic (CPQ pricing, validation, enrichment) with no runtime error, only incorrect behavior.

## 🔧 Solution

- Added `checkCpqAppHandlerHook()` method that queries `{namespace}__CustomClassImplementation__c` for a `CpqAppHandlerHook` record in the source org, cached once per migration run
- Added `case 'Remote Action':` to the `mapElementData` switch that sets `remoteOptions.PreHook = true` and `remoteOptions.PostHook = true` when the hook is registered
- Query failure (permission issues, missing object) gracefully defaults to `false` with a warning log — migration continues without setting hooks rather than aborting
- The hook check runs once in `migrate()` before the element processing loop (not per-element), stored in `this.hasCpqAppHandlerHook`
- The new case mutates the already-parsed `propertySet` object in place; the existing `JSON.stringify` after the switch handles serialization

**Files changed:**
- `src/migration/omniscript.ts`
- `test/migration/omniscript.test.ts` (new)

## 🧪 Testing

Added 5 Mocha test scenarios covering:
1. Hook registered → PreHook=true, PostHook=true set on Remote Action elements
2. No hook registered → remoteOptions unchanged (no keys added)
3. Flags already true + hook registered → values preserved (idempotent)
4. Non-Remote-Action element type → not affected by this change
5. CustomClassImplementation__c query failure → defaults to false, logs warning, migration continues

Static analysis: ESLint PASSED (pre-existing warning in unrelated file only), TypeScript `tsc --noEmit` PASSED

## Impact Assessment

- **Breaking changes:** None — existing migration behavior for all non-Remote-Action element types is unchanged
- **Performance impact:** One additional SOQL query per migration run (cached result), negligible
- **Related WIs:** W-22403162 (this story), TD-0271459 (parent initiative), W-22202779–W-22202839 (Core-side PreHook/PostHook implementation)

### DIFF
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -248,5 +248,6 @@
   "unableToFindActiveOmniscript": "Unable to find active Omniscript \"%s\".",
   "noOmniscriptFoundForOsInstance": "No Omniscript found for OmniscriptInstance \"%s\"",
   "omniscriptNeedsToBeMigrated": "Omniscript \"%s\" needs to be migrated.",
-  "errorDuringSaveForLaterAssessment": "We've encountered errors during the Save For Later assessment: %s"
+  "errorDuringSaveForLaterAssessment": "We've encountered errors during the Save For Later assessment: %s",
+  "prePostHookAutoEnabled": "PreHook/PostHook will be auto-enabled on Remote Action step(s): %s. A hook implementation is registered for the remote class used by these steps."
 }
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -239,6 +239,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     try {
       const exportComponentType = this.getName() as ComponentType;
       const omniscripts = await this.getAllOmniScripts();
+      this.hasCpqAppHandlerHook = await this.checkCpqAppHandlerHook();
 
       if (isStandardDataModelWithMetadataAPIEnabled()) {
         // For the Standard Data Model Orgs, we only need to prepare the storage
@@ -429,6 +430,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     const namespaceErrors: string[] = [];
 
     //const missingRA: string[] = [];
+    const hookEnabledSteps: string[] = [];
 
     // Check for duplicate element names within the same OmniScript
     const elementNames = new Set<string>();
@@ -533,6 +535,9 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
           } else if (status === ApexResolveStatus.NOT_FOUND) {
             namespaceErrors.push(this.messages.getMessage('apexClassNotFound', [className, nameVal]));
           }
+        }
+        if (this.hasCpqAppHandlerHook) {
+          hookEnabledSteps.push(nameVal);
         }
       }
       // To handle radio , multiselect
@@ -761,6 +766,15 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     const uniqueMissingIP = [...new Set(missingIP)];
     const uniqueMissingOS = [...new Set(missingOS)];
 
+    const infos: string[] = [];
+    if (hookEnabledSteps.length > 0) {
+      infos.push(
+        `PreHook/PostHook will be auto-enabled on ${
+          hookEnabledSteps.length
+        } Remote Action step(s): ${hookEnabledSteps.join(', ')}`
+      );
+    }
+
     const result: OSAssessmentInfo = {
       name: recordName,
       id: omniscript['Id'],
@@ -770,7 +784,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       dependenciesOS: uniqueOS,
       dependenciesRemoteAction: uniqueRA,
       dependenciesLWC: uniqueLWC,
-      infos: [],
+      infos: infos,
       warnings: warnings,
       errors: [],
       migrationStatus: assessmentStatus,
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -47,6 +47,7 @@ import { ApexNamespaceRegistry, ApexResolveStatus } from './ApexNamespaceRegistr
 export class OmniScriptMigrationTool extends BaseMigrationTool implements MigrationTool {
   private readonly exportType: OmniScriptExportType;
   private readonly allVersions: boolean;
+  private hasCpqAppHandlerHook: boolean = false;
   private IS_STANDARD_DATA_MODEL: boolean = isStandardDataModel();
   private readonly apexNamespaceRegistry: ApexNamespaceRegistry = ApexNamespaceRegistry.getInstance();
 
@@ -85,6 +86,18 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     // Configure the shared Custom CSS registry. Idempotent — safe to call from
     // both the OS and IP tool instances within a single assess run.
     CustomCssRegistry.getInstance().init(connection, namespace, messages);
+  }
+
+  private async checkCpqAppHandlerHook(): Promise<boolean> {
+    try {
+      const objectName = `${this.namespacePrefix}CustomClassImplementation__c`;
+      const soql = `SELECT Id FROM ${objectName} WHERE Name = 'CpqAppHandlerHook' LIMIT 1`;
+      const result = await this.connection.query(soql);
+      return result.totalSize > 0;
+    } catch (e) {
+      Logger.warn('Unable to query CustomClassImplementation__c for CpqAppHandlerHook: ' + e);
+      return false;
+    }
   }
 
   getName(
@@ -918,6 +931,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
   async migrate(): Promise<MigrationResult[]> {
     // Get All Records from OmniScript__c (IP & OS Parent Records)
     const omniscripts = await this.getAllOmniScripts();
+    this.hasCpqAppHandlerHook = await this.checkCpqAppHandlerHook();
 
     if (isStandardDataModelWithMetadataAPIEnabled()) {
       return this.handleMigrationForStdDataModelOrgsWithMetadataAPIEnabled(omniscripts);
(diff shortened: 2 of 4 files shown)


twu-sf (approved):

src/migration/omniscript.ts
```diff
@@ -736,6 +754,15 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     const uniqueMissingIP = [...new Set(missingIP)];
     const uniqueMissingOS = [...new Set(missingOS)];
 
+    const infos: string[] = [];
+    if (hookEnabledSteps.length > 0) {
+      infos.push(
+        `PreHook/PostHook will be auto-enabled on ${
```
rajender-kumar-salesforce: UX labels approval required. Also use string constant for it and read using messages.getMessage() . Check on UI how its being shown currently.

rajender-kumar-salesforce: I think it make sense to put this as warnings for each Integration procedure with step.

src/migration/omniscript.ts
```diff
@@ -745,7 +772,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       dependenciesOS: uniqueOS,
       dependenciesRemoteAction: uniqueRA,
       dependenciesLWC: uniqueLWC,
-      infos: [],
+      infos: infos,
```
rajender-kumar-salesforce: processOmniComponents is dropping infos and assigning to empty array .

src/migration/omniscript.ts
```diff
@@ -226,6 +239,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     try {
       const exportComponentType = this.getName() as ComponentType;
       const omniscripts = await this.getAllOmniScripts();
+      this.hasCpqAppHandlerHook = await this.checkCpqAppHandlerHook();
```
rajender-kumar-salesforce: why we are running checks  for Omniscript. Hooks are IP specific.

rajender-kumar-salesforce: Applicable for both Asses and Migration.

src/migration/omniscript.ts
```diff
@@ -2413,6 +2441,13 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     if (propSetMap.remoteClass) {
       propSetMap.remoteClass = this.apexNamespaceRegistry.getQualifiedClassName(propSetMap.remoteClass);
     }
+
+    if (this.hasCpqAppHandlerHook) {
```
rajender-kumar-salesforce: We should not set these flags for every IP even mapping exists. These mappings may not be used in IP.
1. Check IP has Step Name -> Remote Action.
2. Check if Mapping Exists for ClassName -> (Ex: CpqAppHandler ) (can be added &b read from cached to avoid multiple SOQL query).
3. if yes enable hooks.
4. Add warning to specific IP, step name that hooks are enabled for both assess and migrate mode.

src/migration/omniscript.ts
adamespii: Addressed — moved the string to messages.getMessage() in both migrate.json and assess.json (key: `prePostHookAutoEnabled`). The current wording is a placeholder pending UX approval. Who would be the best person to reach out to for UX label sign-off on this?

adamespii: Addressed in b7eaa5e — moved from `infos[]` to `warnings[]`, with a per-step warning for each affected Remote Action element on the Integration Procedure.

src/migration/omniscript.ts
adamespii: Good catch — this is now moot since we moved to `warnings[]` which are already forwarded through `processOmniComponents` into the assessment result. No more dropped data.

src/migration/omniscript.ts
adamespii: Good point — addressed in b7eaa5e. The hook warning is now gated to `omniProcessType === 'Integration Procedure'` only. OmniScripts won't get the hook check or warning.

adamespii: Confirmed — the hook check now runs in both `assess()` and `migrate()` paths. Both are covered.

src/migration/omniscript.ts
adamespii: Addressed in b7eaa5e — fully reworked per your suggestions:

1. No longer blanket-enables on all Remote Action steps.
2. `loadHookRegistrations()` queries all `*Hook` entries from `CustomClassImplementation__c` (cached once per run).
3. `hasHookForClass(remoteClass)` extracts the simple class name and checks if it has a matching hook registration.
4. Only steps whose `remoteClass` has a registered hook get PreHook/PostHook enabled.
5. Per-step warning added in both assess and migrate mode (via `messages.getMessage('prePostHookAutoEnabled', [stepName])`).

src/migration/omniscript.ts
```diff
@@ -736,6 +769,12 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     const uniqueMissingIP = [...new Set(missingIP)];
     const uniqueMissingOS = [...new Set(missingOS)];
 
+    if (hookEnabledSteps.length > 0 && omniProcessType === 'Integration Procedure') {
+      for (const stepName of hookEnabledSteps) {
+        warnings.push(this.messages.getMessage('prePostHookAutoEnabled', [stepName]));
```
rajender-kumar-salesforce: Instead of 10 warning for 10 different steps. We can have single warning for all steps to avoid unnecessary cluttered texts

src/migration/omniscript.ts
```diff
@@ -226,6 +254,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
     try {
       const exportComponentType = this.getName() as ComponentType;
       const omniscripts = await this.getAllOmniScripts();
+      this.hookRegisteredClasses = await this.loadHookRegistrations();
```
rajender-kumar-salesforce: Does it run for Omniscript ?

messages/assess.json
```diff
@@ -239,5 +239,6 @@
...
   "apexRemoteDatasourceNamespaceWarning": "ApexRemote datasource: remoteClass will be updated from \"%s\" to \"%s\"",
   "apexClassNotFound": "Apex class \"%s\" referenced in \"%s\" does not exist in the org",
-  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>"
+  "cleanupMetadataTablesRequired": "Omnistudio configuration tables contain records. Back up your Omnistudio component table data, and manually clean up the Omnistudio configuration tables. <a href='https://help.salesforce.com/s/articleView?id=xcloud.os_enable_omnistudio_metadata_api_support.htm&type=5' target='_blank'>Learn more about cleaning up tables in Salesforce Help</a>",
+  "prePostHookAutoEnabled": "PreHook/PostHook will be auto-enabled on Remote Action step '%s' during migration because a hook implementation is registered for its remote class."
```
rajender-kumar-salesforce: Take UX Approvals for all new labels.

src/migration/omniscript.ts
```diff
@@ -906,6 +945,7 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
   async migrate(): Promise<MigrationResult[]> {
     // Get All Records from OmniScript__c (IP & OS Parent Records)
     const omniscripts = await this.getAllOmniScripts();
+    this.hookRegisteredClasses = await this.loadHookRegistrations();
```
rajender-kumar-salesforce: Does it run for Omniscript ?

src/migration/omniscript.ts
rajender-kumar-salesforce: @adamespii for UX reach out to Rasmi Radhakrishnan

akshatsinha-sf (approved):

src/migration/omniscript.ts
adamespii: Addressed — consolidated into a single warning per IP that lists all affected step names (e.g. "PreHook/PostHook will be auto-enabled on Remote Action step(s): Step1, Step2..."). No more per-step clutter.

src/migration/omniscript.ts
adamespii: Good catch — added a guard: `if (this.exportType !== OmniScriptExportType.OS)`. The hook registration query is now skipped entirely for OmniScript-only runs. It only fires for IP or All export types.

messages/assess.json
adamespii: Acknowledged — will reach out to Rasmi Radhakrishnan for UX approval on the label wording. Current text is a placeholder.

src/migration/omniscript.ts
adamespii: Same as above — added guard `if (this.exportType !== OmniScriptExportType.OS)` in `migrate()` too. Hook query skipped for OmniScript-only migration runs.

src/migration/omniscript.ts
adamespii: Thanks Rajender — will reach out to Rasmi Radhakrishnan for UX sign-off on the warning text.

rajender-kumar-salesforce (approved):

PR #493, opened 2026-05-22
@W-22334153 Adding fixes for navigate action in Omniscript

sf-pallavi-das:
### What does this PR do?

- Added processNavigateAction handler in that resolves Type/Sub Type/Language from the legacy LWC OmniScript reference using the name registry and clears the LWC name field.
- Rewrites OmniScript Prefill keys from the managed-package c__ namespace to the standard runtime omniscript__ namespace (only at key boundaries, not inside values).
- Wired the new Navigate Action constant into the dispatcher and added 6 unit tests in covering match/no-match, prefill rewriting, and dispatcher integration.

### DIFF
--- a/src/migration/omniscript.ts
+++ b/src/migration/omniscript.ts
@@ -2129,6 +2129,9 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
       case Constants.RemoteAction:
         this.processRemoteAction(propSet);
         break;
+      case Constants.NavigateAction:
+        this.processNavigateAction(propSet);
+        break;
       default:
         // Handle other element types if needed
         break;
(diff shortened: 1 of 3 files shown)


shaurabh-tiwari-git (approved):
LGTM

mdmehran-qureshi (approved):
.

src/migration/omniscript.ts
```diff
@@ -2250,6 +2253,46 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+   * to the standard runtime's `omniscript__` namespace.
+   * @param propSetMap Property set map from the element
+   */
+  private processNavigateAction(propSetMap: any): void {
+    const lwcRef: string = typeof propSetMap['LWC OmniScript'] === 'string' ? propSetMap['LWC OmniScript'] : '';
+    if (lwcRef) {
+      const stripped = lwcRef.replace(/^c:/, '').replace(/^c__/, '').toLowerCase();
+      const match = this.nameRegistry.getOmniScriptMappingKeys().find((key) => {
+        const parts = key.split('_');
+        if (parts.length < 2) return false;
```
mdmehran-qureshi, line 2395: **Suggestion:** `Array.find()` returns the first match. If two registry keys normalize to the same lowercase string (e.g., `Customer_Profile_English` vs `CustomerProfile_English` after cleaning), the match is non-deterministic. Consider logging a warning if multiple candidates match, or documenting that this is expected to be unique.

src/migration/omniscript.ts
```diff
@@ -2250,6 +2253,46 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+  /**
+   * Processes Navigate Action elements so the migrated parent OmniScript can launch its child
+   * under the standard runtime: resolves Type/Sub Type/Language from the legacy `LWC OmniScript`
+   * reference and rewrites `OmniScript Prefill` keys from the managed-package `c__` namespace
+   * to the standard runtime's `omniscript__` namespace.
+   * @param propSetMap Property set map from the element
+   */
+  private processNavigateAction(propSetMap: any): void {
+    const lwcRef: string = typeof propSetMap['LWC OmniScript'] === 'string' ? propSetMap['LWC OmniScript'] : '';
+    if (lwcRef) {
```
mdmehran-qureshi, line 2386: **Nit:** Only `c:` and `c__` prefixes are stripped here. In some LWC contexts, `c/` is also used as a namespace separator. If that's not a valid scenario for the `LWC OmniScript` field value, this is fine — but worth confirming or adding a comment noting the assumption.

src/migration/omniscript.ts
```diff
@@ -2250,6 +2253,46 @@ export class OmniScriptMigrationTool extends BaseMigrationTool implements Migrat
...
+        const type = parts[0];
+        const subType = parts[1];
+        const language = parts[2] || 'English';
+        const candidate = `${this.cleanName(type)}${this.cleanName(subType)}${language}`.toLowerCase();
+        return candidate === stripped;
+      });
+
+      if (match) {
+        const cleanedFullName = this.nameRegistry.getCleanedName(match, 'OmniScript');
+        const cleanedParts = cleanedFullName.split('_');
```
mdmehran-qureshi, line 2414: **Suggestion:** If `getCleanedName()` returns a single-segment string (no `_` separator), this guard silently skips setting Type/Sub Type — but the LWC field is still cleared on line 2285. This could leave the element in a broken state. Consider adding a warning log inside an `else` branch here.

test/migration/omniscript-content-processing.test.ts
```diff
@@ -308,6 +308,91 @@ describe('OmniScript Content Processing - Comprehensive Tests', () => {
...
+      expect(propSetMap.Type).to.equal('Customer');
+      expect(propSetMap['Sub Type']).to.equal('ProfileCleaned');
+      expect(propSetMap.Language).to.equal('English');
+      expect(propSetMap['LWC OmniScript']).to.equal('');
+      expect(propSetMap['Page Reference Type']).to.equal('Vlocity OmniScript');
+    });
+
+    it('should rewrite c__ prefill keys to omniscript__ at key boundaries', () => {
+      const propSetMap: any = {
+        'LWC OmniScript': '',
```
mdmehran-qureshi: **Suggestion:** Consider adding a test with the `c__` prefix variant (e.g., `c__customerprofileEnglish`) to confirm both regex branches in the stripping logic work correctly.

sf-kishore-kurri (approved):
Can you add the video recording of the working use case post migration, changes LGTM

test/migration/omniscript-content-processing.test.ts
sf-pallavi-das: addressed

src/migration/omniscript.ts
sf-pallavi-das: addressed

src/migration/omniscript.ts
sf-pallavi-das: addressed

src/migration/omniscript.ts
sf-pallavi-das: The current regex /^[^:]+:/ (line 2387) strips any namespace ending in : — covers c:, mns:, etc. The c/ form only appears in ES-module imports (import X from 'c/foo'); it never appears in propertySet targetLWC values, which use the component-tag form

PR #496, opened 2026-05-25
@W-20048799 - Identification of action keyword during assessment

shaurabh-tiwari-git:

### DIFF
--- a/messages/assess.json
+++ b/messages/assess.json
@@ -99,7 +99,7 @@
   "cardNameChangeMessage": "The Flexcard name will be changed from %s to %s to adhere to the API naming standards.",
   "authordNameChangeMessage": "The Flexcard author name will be changed from %s to %s to adhere to the naming rules",
   "cardLWCNameChangeMessage": "The Flexcard generated LWC name will be changed from %s to %s to align with Flexcard name change.",
-  "customLwcReservedActionKey": "Custom LWC \"%s\" defines a reserved property key \"action\" which is not supported by the core runtime. Manual intervention required to remove or rename this key before migration.",
+  "customLwcReservedActionKey": "Custom LWC \"%s\" defines a reserved property key \"action\" which is not supported by the core runtime. Manual intervention required to rename this key to a non reserved key before migration.",
   "webPageOmniScriptNavigationDetected": "Detected Omniscript navigation in web page action URL: %s.\n\nIt will be replaced with: %s",
   "webPageOmniScriptUrlMalformedToken": "Detected Omniscript navigation in web page action URL with malformed fragment token(s) [%s]. The token(s) couldn't be URL-decoded and were kept as-is.\nURL: %s.\n\nIt will be replaced with: %s.\nVerify that the rewritten URL points to the intended Omniscript before using this Flexcard in the destination org.",
   "omniScriptNameChangeMessage": "The Omniscript reference name %s will be changed to %s during migration.",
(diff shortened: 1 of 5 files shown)


mdmehran-qureshi (approved):

sf-pallavi-das (approved):

src/migration/flexcard.ts
```diff
@@ -631,6 +631,16 @@ export class CardMigrationTool extends BaseMigrationTool implements MigrationToo
           flexCardAssessmentInfo.dependenciesLWC.push(customLwcName);
         }
       }
+
+      // "action" is a reserved keyword in the core runtime; flag any Custom LWC
+      // that defines it as a property so the user removes/renames it pre-migration.
+      if (Object.prototype.hasOwnProperty.call(component.property, 'action')) {
+        const lwcLabel = component.property.customlwcname || component.elementLabel || Constants.CustomLwc;
```
sf-kishore-kurri, line 646: @shaurabh-tiwari-git As discussed over call, connect with Swati CX and add all the reserve keywords check with Flexcard custom attributes

messages/assess.json
```diff
@@ -99,6 +99,7 @@
   "cardNameChangeMessage": "The Flexcard name will be changed from %s to %s to adhere to the API naming standards.",
   "authordNameChangeMessage": "The Flexcard author name will be changed from %s to %s to adhere to the naming rules",
   "cardLWCNameChangeMessage": "The Flexcard generated LWC name will be changed from %s to %s to align with Flexcard name change.",
+  "customLwcReservedActionKey": "Custom LWC \"%s\" defines a reserved property key \"action\" which is not supported by the core runtime. Manual intervention required to remove or rename this key before migration.",
```
sf-kishore-kurri: Message is not correct - remove is not an option which will impact the functionality so it should be renamed not using reserve keywords which will be shared as documentation link

src/utils/constants/documentRegistry.ts
```diff
@@ -32,4 +32,6 @@ export const documentRegistry = {
   customLabelMigrationErrorMessage:
     'https://help.salesforce.com/s/articleView?id=xcloud.os_migrate_oma_prereq.htm&type=5',
   corruptedParentChildLevel: 'https://help.salesforce.com/s/articleView?id=xcloud.os_version_omniscripts.htm&type=5',
+  customLwcReservedActionKey:
+    'https://help.salesforce.com/s/articleView?id=xcloud.os_embed_a_custom_lwc_inside_a_flexcard.htm&type=5',
```
sf-kishore-kurri: replace the link with CX created link which has clear steps on what should be done with this manual step

src/migration/flexcard.ts
shaurabh-tiwari-git: I've tried with other reserved keywords but they are not causing any issue in our case. They might be reserved for other places in the Flexcard but for our case only action keyword is failing in core runtime. Also, the PR by Adam in core for this issue (https://gitcore.soma.salesforce.com/core-2206/core-264-public/pull/2750) only addresses the "action" keyword.

messages/assess.json
shaurabh-tiwari-git: Addressed

src/utils/constants/documentRegistry.ts
shaurabh-tiwari-git: Addressed

PR #499, opened 2026-05-27
feat: @W-22534088 Migration phase for Omniscript saved session migration

twu-sf:
### What does this PR do?
Adds the migration logic for Omniscript Saved Sessions into the migration tool.
The migrate() function will do the following 
1. Check if the org has the custom field : PackageSavedSessionId__c for OmniscriptSavedSession
2. Query for all of the package Omniscript Saved Sessions (OmniscriptInstance__c)
3. Query for all "migrated" Omni Processes with the same type, subtype and language as the Saved Session counterparts.
4. Do some error handling, checks for missing fields, replace/ update fields.
5. Create the new records for core Omniscript Saved Sessions (OmniscriptSavedSession)
6. Download the old attachments associated with the package Omniscript Saved Sessions (OmniscriptInstance__c)
7. Do some error handling, checks for missing fields, replace/ update fields.
8. Upload the attachments to associate with the new core Omniscript Saved Sessions (OmniscriptSavedSession)
9. More error handling
10. Send to have the migration report be generated and displayed.


akshatsinha-sf (approved):

src/migration/omniscriptInstance.ts
```diff
@@ -117,29 +125,75 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+
+    const originalOsInstanceRecords = new Map<string, unknown>();
+    const osInstanceUploadInfo = new Map<string, UploadRecordResult>();
+
+    try {
+      // verify custom field PackageSavedSessionId__c exists in OmniScriptSavedSession
+      const hasCustomField = await this.hasCustomFieldPackageSavedSessionId();
+      if (!hasCustomField) {
+        Logger.error(
+          'Missing custom field. Please add custom field PackageSavedSessionId__c to OmniscriptSavedSession. Please read docs.'
```
aakkineni-gh: Can we create final constants for this?

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+          Logger.warn(errorMsg);
+          errors.push(errorMsg);
+        }
+      }
+
+      Logger.log(`Uploaded ${attachmentsUploaded}/${attachments.length} attachments for ${osInstanceName}`);
+    }
+
+    return {
+      attachmentUploadSuccess: attachmentsUploaded === 3,
```
aakkineni-gh, line 641: Is this always 3 attachments? Can we instead just check the length? 
`attachmentUploadSuccess: attachments.length`

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+      await this.migrateCreateOmniscriptSavedSession(
+        osInstanceId,
+        osInstanceName,
+        osType,
+        osSubType,
+        osLanguage,
+        savedSessionData
+      );
+    if (omniscriptSavedSessionResult?.errors?.length > 0) {
+      errors.concat(omniscriptSavedSessionResult.errors);
```
aakkineni-gh: Do we need to do: `errors.push(...omniscriptSavedSessionResult.errors)`

I think concat is immutable and it creates a new array and when unassigned new array is discarded.

Also we can probably verify via any UT

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+      // Transform / replace references to managed package in attachments Body  : OmniscriptFullJSON.json
+
+      // Upload Attachment with Saved Session's Id as ParentId
+      const uploadAttachmentResult: MigrateUploadAttachmentsResult = await this.migrateUploadAttachments(
+        attachmentBodyData,
+        newSavedSessionId,
+        osInstanceName
+      );
+      if (uploadAttachmentResult?.errors?.length > 0) {
+        errors.concat(uploadAttachmentResult.errors);
```
aakkineni-gh, line 481: Similar as above comment - pls verify errors are actually appended.

src/migration/omniscriptInstance.ts
```diff
@@ -374,11 +811,83 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+   * @returns Array of downloaded attachment data with bodies and any errors
+   */
+  private async downloadAttachments(attachments: AnyJson[]): Promise<AttachmentDownloadResult[]> {
+    const results: AttachmentDownloadResult[] = [];
+
+    const errors: string[] = [];
+
+    for (const attachment of attachments) {
+      if (!attachment) {
+        errors.push('Failed to download attachment');
```
aakkineni-gh: We can probably be more specific something like
 `errors.push(`Attachment at index ${attachments.indexOf(attachment)} is null/undefined, skipping`);
}`

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+
+      if (!uploadResult.success || uploadResult.hasErrors) {
+        Logger.error(`Failed to upload OmniScriptSavedSession for ${osInstanceName} ${JSON.stringify(uploadResult)}`);
+        errors.push(JSON.stringify(uploadResult.errors));
+      }
+
+      Logger.logVerbose(`Successfully updated ResumeUrls for OmniScriptSavedSession: ${uploadResult.id}`);
+    } catch (error) {
+      const errorMsg = error instanceof Error ? error.message : String(error);
+      Logger.error(`Error uploading OmniScriptSavedSession for ${osInstanceName}: ${errorMsg}`);
```
aakkineni-gh: This update and above line #532 looks same do we want it to be unique?

src/migration/omniscriptInstance.ts
```diff
@@ -374,11 +811,83 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+   */
+  private async downloadAttachments(attachments: AnyJson[]): Promise<AttachmentDownloadResult[]> {
+    const results: AttachmentDownloadResult[] = [];
+
+    const errors: string[] = [];
+
+    for (const attachment of attachments) {
+      if (!attachment) {
+        errors.push('Failed to download attachment');
+        break;
```
aakkineni-gh: Wondering on this scenario further: what will be the state in `OmniscriptSavedSession` when these error scenarios occur for a particular atachment.

Say Like attachment 1 went through and error occurred in attached 2 - we stop processing here. How do we clean up/roll back or retry/resume again?

Might be relevant for other errors as well.

twu-sf: For this scenario, we do not rollback, instead, we report an error in the terminal / dashboard that the attachment failed.

twu-sf: We could consider adding retries for failed uploads or consider adding rollback logic.

rajender-kumar-salesforce: Currently, we break when encountering a null attachment. Should we instead skip null entries and continue processing the remaining valid attachments? Another approach would be to pre-filter the attachment list and either upload all valid attachments or none at all.

src/migration/omniscriptInstance.ts
```diff
@@ -374,11 +811,83 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+   * Download attachment bodies from Salesforce
+   *
+   * @param connection - Salesforce connection instance
+   * @param attachmentBodyPaths - Array of attachment body URLs (e.g., /services/data/v67.0/sobjects/Attachment/00PSB000003DJG12AO/Body)
+   * @returns Array of downloaded attachment data with bodies and any errors
+   */
+  private async downloadAttachments(attachments: AnyJson[]): Promise<AttachmentDownloadResult[]> {
+    const results: AttachmentDownloadResult[] = [];
+
+    const errors: string[] = [];
```
rajender-kumar-salesforce: Move this inside loop per attachement

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+        recordId
+      );
+    }
+
+    try {
+      uploadResult = await NetUtils.updateOne(
+        this.connection,
+        Constants.OmniScriptSavedSessionObjectName,
+        osInstanceId,
+        recordId,
```
rajender-kumar-salesforce: Can it be empty when creation of record fails ?

messages/migrate.json
rajender-kumar-salesforce: Missing --only=sfl in CLI Examples

src/migration/omniscriptInstance.ts
```diff
@@ -282,12 +355,345 @@ export class OmniScriptInstanceMigrationTool extends BaseMigrationTool implement
...
+
+      return uploadedRecord;
+    }
+
+    // Extract OmniProcess Id from omniProcessSet
+    const uniqueOmniProcessString = osType + osSubType + osLanguage;
+    const targetOmniProcessId = omniProcessMap.get(uniqueOmniProcessString);
+
+    if (!targetOmniProcessId) {
+      errors.push(`No matching OmniProcess found for ${osType}_${osSubType}_${osLanguage}`);
```
rajender-kumar-salesforce: Move to JSON Skipping attachment upload, Failed to download attachment. Please take UX Approval for these labels

src/migration/omniscriptInstance.ts
twu-sf: Moved this code block so that it will have the recordId checked before performing update.

src/migration/omniscriptInstance.ts
twu-sf: Updated.

src/migration/omniscriptInstance.ts
twu-sf: There is check before the actual function call so either all attachments are uploaded or none. I will remove this additional break as it is not necessary.

src/migration/omniscriptInstance.ts
twu-sf: Messages have been cleaned up. Thanks!

src/migration/omniscriptInstance.ts
twu-sf: Removed the error message as it is redundant.

src/migration/omniscriptInstance.ts
twu-sf: Updated.

src/migration/omniscriptInstance.ts
twu-sf: Updated

src/migration/omniscriptInstance.ts
twu-sf: attachmentsUploaded gets incremented with each successful upload

src/migration/omniscriptInstance.ts
twu-sf: yes.

twu-sf:
The two failing tests were due to a specific chrome version not found. All unit tests on linux and windows passed for different versions of chrome.

PR #503, opened 2026-06-11
fix: @W-22360038 Fixing npm audit vulnerabilties

sf-pallavi-das:
## Summary

Fixes 126 of 137 npm audit vulnerabilities (137 → 11; 0 critical/high remaining). All remaining advisories are dev-only (`diff`, `uuid` in `mocha`/`nyc`/`ts-sinon` chains) and don't ship in `/lib`.

**Severity breakdown:**

|          | Before | After |
| -------- | ------ | ----- |
| Critical | 4      | 0     |
| High     | 71     | 0     |
| Moderate | 50     | 2     |
| Low      | 12     | 9     |
| **Total**| **137**| **11**|

## Changes

- **Direct dep bumps** (patch versions): `@xmldom/xmldom` 0.9.8 → 0.9.10 (5 high CVEs in XML parsing path), `diff` 5.1.0 → 5.2.2
- **Yarn `resolutions` added** for: `handlebars` ^4.7.9 (2 critical), `basic-ftp` ^5.3.1 (critical via puppeteer), `lodash` ^4.18.0, `serialize-javascript` ^7.0.5, `flatted` ^3.4.2, `ws` ^8.20.1, `yaml` ^1.10.3, `fast-uri` ^3.1.2, `ip-address` ^10.1.1, `picomatch` ^4.0.4
- **Yarn `resolutions` tightened**: `minimatch` ^3.0.5 → ^3.1.4, `brace-expansion` ^2.0.2 → ^2.0.3, `tmp` ^0.2.4 → ^0.2.6
- **`engines.node`**: `>=12.0.0` → `>=20.0.0` — the existing `>=12` was already inaccurate (`@salesforce/core`, `@salesforce/sf-plugins-core`, `jsdom`, `puppeteer` all require ≥18; Node 18 is EOL since April 2025). Bumping unlocks `serialize-javascript@7` cleanly (requires Node 20). Aligns with current LTS.
- **README.md**: updated Node version requirement docs from 18 → 20

## Notes for reviewers

- `serialize-javascript@7` requires Node ≥20, which is why `engines` had to bump. Node 18 was already EOL when this PR was authored.
- Remaining 11 advisories all live in dev tooling (`mocha@8`, `nyc@15`, `ts-sinon@1`, `ts-node@10`) and require dev-tooling modernization (Phase 2: `nyc → c8`, `mocha 8 → 10`, etc.) — out of scope for this audit-fix PR.
- The yarn `resolutions` block already had a strong precedent (~25 entries) — this PR continues that pattern.


sf-kishore-kurri:
Security Review — npm audit fix (supply-chain lens)
@sf-pallavi-das — thanks for tackling this. I validated the lockfile against the live npm registry (version existence, integrity-hash authenticity, resolution effectiveness, and resolutions-vs-dependencies placement). The fix is fundamentally sound — the runtime-facing @xmldom/xmldom bump is correct, transitive pins land in resolutions, and every lockfile integrity hash I checked matches the registry byte-for-byte (no tampering). Two pins overshoot, though, and one forces an unnecessary consumer-facing breaking change. Recommending changes before merge.

✅ What checks out
All 15 pinned versions exist on npm; yarn.lock integrity verified for the 7 highest-risk packages (xmldom, handlebars, serialize-javascript, basic-ftp, diff, picomatch, lodash).
Resolutions actually moved the lockfile (no silent no-ops); transitive pins correctly in resolutions, direct deps in dependencies.
@xmldom/xmldom 0.9.8 → 0.9.10 (ships in /lib): correct, same-major, addresses the XML-parse highs. 👍
handlebars ^4.7.9, basic-ftp ^5.3.1, and the minor/patch bumps (diff, lodash, ws, yaml, flatted, fast-uri, ip-address, brace-expansion, tmp): same-major, low risk.
Deferring the 11 residual dev-only advisories to a Phase-2 tooling refresh is a reasonable accept-and-defer.
🟠 1. serialize-javascript ^7.0.5 overshoots and forces Node 20 (highest priority)
The XSS advisory (GHSA-h9rv-jmmf-4pgx) was fixed in 6.0.2; this pins 7.0.5 (two majors past the fix). v7 is the only reason engines.node jumps to >=20 (v6 has no engines constraint). Since serialize-javascript is dev-only here (mocha/nyc chain, not in /lib), a dev-only moderate advisory is raising the published Node floor for every end user of the plugin.

Suggested: pin serialize-javascript: ^6.0.2 and keep engines.node at >=18. If the Node-20 move is intentional, split it into its own PR with its own rationale, and please confirm Node 20 ≥ the sf CLI supported-runtime matrix (an engines bump can break plugin install on older CLIs).

🟠 2. picomatch ^4.0.4 collapses three majors into one
Before this PR, picomatch correctly resolved to three coexisting versions (2.3.1 / 3.0.1 / 4.0.3). This forces all consumers (^2.x, ^3.x, ^4.x) to 4.0.4 — a two-major jump for 2.x/3.x consumers (chokidar/readdirp/fast-glob/tinyglobby). I couldn't find a high/critical advisory that requires picomatch 4.

Suggested: drop the picomatch resolution unless a specific advisory drives it; if one does, scope it to the vulnerable major rather than a blanket ^4.0.4.

🟡 3. Pre-existing over-broad minimatch pin (not introduced here)
The global minimatch entry forces consumers requesting ^5, ^9, ^10 down to 3.1.5 (a force-downgrade). 3.1.5 is patched for CVE-2022-3517 so no known CVE, but it's a correctness/maintainability liability. Out of scope to fix here — flagging so it isn't read as "all clear."

🟢 Open items to make the fix auditable
Attach yarn audit output for the branch proving 0 critical/high — the 137→11 numbers are currently asserted, not reproduced in-PR.
Add a pkg → advisory ID → fixed-in table so each pin's justification is auditable (this is what surfaces overshoots like #1/#2).
Consider exact pins instead of ^ ranges for security resolutions, for reproducibility.
Recommended before merge
serialize-javascript: ^6.0.2 + revert engines.node to >=18 (or split the Node-20 bump out).
Drop/scope the picomatch resolution.
Attach yarn audit proof + advisory mapping table.
Note: validated registry existence, lockfile integrity authenticity, resolution effectiveness, and block placement directly; did not run yarn install && yarn audit on the branch, so the 0-critical/high claim rests on the reported numbers plus version checks.

shaurabh-tiwari-git (approved):
LGTM

yarn.lock
```diff
@@ -8046,10 +8078,10 @@ write-json-file@^4.1.1:
...
     write-file-atomic "^3.0.0"
 
-ws@^8.18.0, ws@^8.18.3:
-  version "8.18.3"
-  resolved "https://registry.yarnpkg.com/ws/-/ws-8.18.3.tgz#b56b88abffde62791c639170400c93dcb0c95472"
-  integrity sha512-PEIGCY5tSlUt50cqyMXfCzX+oOPqN0vuGqWzbcJ2xvnkzkq46oOpz7dQaTDBdfICb4N14+GARUDw2XV2N4tvzg==
+ws@^8.18.0, ws@^8.18.3, ws@^8.20.1:
+  version "8.21.0"
+  resolved "https://registry.yarnpkg.com/ws/-/ws-8.21.0.tgz#012e413fc07429945121b0c153158c4343086951"
+  integrity sha512-Vsp28b7DRcimFQvrqu2Wek3z1iYxDCWqHYB8Qsnk/S4RfaCQzPGPyBNuVjJV3cd6UiKtUtp6sNM77gWvzcCH+g==
```
sf-kishore-kurri, line 8012: @sf-pallavi-das Do we. check-in lock files? @snehaljha-sf

mdmehran-qureshi (approved):
lgtm

yarn.lock
sf-pallavi-das: As per best practices, we should check in lock files.

PR #504, opened 2026-06-18
fix(dataraptor): set IsMigrated=true for all migrated DataMa…

sf-vijaychetan-pabbisetty:
…ppers

### DIFF
--- a/src/migration/dataraptor.ts
+++ b/src/migration/dataraptor.ts
@@ -550,7 +550,7 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
 
     mappedObject['Name'] = this.cleanName(mappedObject['Name']);
     mappedObject['IsActive'] = true;
-    mappedObject['IsEmptyInputAllowed'] = true;
+    mappedObject['IsMigrated'] = true; 
 
     // BATCH framework requires that each record has an "attributes" property
     mappedObject['attributes'] = {
--- a/src/migration/dataraptor.ts
+++ b/src/migration/dataraptor.ts
@@ -550,6 +550,7 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
 
     mappedObject['Name'] = this.cleanName(mappedObject['Name']);
     mappedObject['IsActive'] = true;
+    mappedObject['IsEmptyInputAllowed'] = true;
 
     // BATCH framework requires that each record has an "attributes" property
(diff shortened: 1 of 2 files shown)


src/migration/dataraptor.ts
```diff
@@ -550,6 +550,7 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
 
     mappedObject['Name'] = this.cleanName(mappedObject['Name']);
     mappedObject['IsActive'] = true;
+    mappedObject['IsEmptyInputAllowed'] = true;
```
sf-kishore-kurri: My only concern with this change is, when we merge this change new OMA plugin version is generated and any customer using this version of plugin will create problem, if schema doesn't have the column exists

We need to figure out how to bind the change with coreapp release version?

sf-kishore-kurri (approved):
Approved

src/migration/dataraptor.ts
```diff
@@ -550,7 +550,12 @@ export class DataRaptorMigrationTool extends BaseMigrationTool implements Migrat
 
     mappedObject['Name'] = this.cleanName(mappedObject['Name']);
     mappedObject['IsActive'] = true;
-
+    // IsMigrated field only exists in core 264+ (API v68.0).
+    // Skip on older orgs to avoid "No such column" errors.
+    const apiVersion = parseFloat((this.connection as any).getApiVersion?.() ?? '0');
+    if (apiVersion >= 68.0) {
```
sf-pranav-varshney, line 556: I'm not sure how the OMA tool evaluates the API version. Just a thought, could there be scenarios where the customer is using a lower API version? If so, is that an expected use case, and does the tool handle it appropriately?

sf-pranav-varshney (approved):

src/migration/dataraptor.ts
sf-vijaychetan-pabbisetty: Yes, that's the exact scenario the gate handles — if the customer is on an API version below 68.0 (Core 264), apiVersion >= 68.0 evaluates to false and we simply skip setting IsMigrated, so the DataMapper record is created without that field. This avoids the INVALID_FIELD: No such column 'IsMigrated' error on pre-264 orgs where the column doesn't exist yet. We validated this against both a 262 org (gate skips it, migration succeeds) and a 264 org (gate sets it, migration succeeds)

PR #506, opened 2026-06-26
@W-23018311 - Skip package lwc utilities which are not available on core

shaurabh-tiwari-git:

### DIFF
--- a/src/utils/lwcparser/LwcPackageUtilityRegistry.ts
+++ b/src/utils/lwcparser/LwcPackageUtilityRegistry.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * Licensed under the BSD 3-Clause license.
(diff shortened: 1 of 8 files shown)


src/utils/lwcparser/LwcPackageUtilityRegistry.ts
```diff
@@ -0,0 +1,81 @@
...
+  private static instance: LwcPackageUtilityRegistry;
+
+  // Sourced from github.com/sf-industries/via_ins@master:lwcprojects/base.
+  private static readonly SEED_UTILITY_COMPONENTS: readonly string[] = [
+    'insAccordion',
+    'insAccordionSection',
+    'insAttribute',
+    'insAttributeCategory',
+    'insAttributeCategoryList',
+    'insChart',
```
sf-kishore-kurri, line 20: Let's not add this exclusion with INS files without discussion with vertical team

sf-kishore-kurri: First let's start the discussion with Insurance vertical to understand the plan and then fix it accorindgly

PR #509, opened 2026-07-08
fix: @W-22959073: auto-enable Pre/Post hooks for delegated classes and org-owned namespaces

sf-IrfanAmeeer-Shaik:
### What does this PR do?
Pre/Post hook auto-enable was only working when remoteClass was exactly CpqAppHandler. Turns out a lot of Remote Actions set remoteClass to the implementation class and put the actual handler in remoteMethod instead (e.g. CpqAppHandler-getCartsItems), so those were getting skipped. Added a fallback to check remoteMethod too, and also look at InterfaceImplementation__c for hook registrations, not just CustomClassImplementation__c.

Added tests 

W-22959073

### DIFF
--- a/src/utils/orgUtils/index.ts
+++ b/src/utils/orgUtils/index.ts
@@ -20,6 +20,7 @@ interface InstalledPackage {
 interface OrgDetails {
   Name: string;
   Id: string;
+  NamespacePrefix?: string;
 }
 
 export interface OmnistudioOrgDetails {
@@ -31,6 +32,7 @@ export interface OmnistudioOrgDetails {
   rollbackFlags?: string[];
   isFoundationPackage: boolean;
   isOmnistudioMetadataAPIEnabled: boolean;
+  isOrgOwnedNamespace?: boolean;
 }
 
 export interface PackageDetail {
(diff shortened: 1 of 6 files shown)


src/utils/orgUtils/index.ts
```diff
@@ -357,16 +361,27 @@ export class OrgUtils {
...
-        orgDetails: orgDetails[0],
-        dataModel: undefined,
-        hasValidNamespace: false,
-        isFoundationPackage: false,
-        isOmnistudioMetadataAPIEnabled: false,
-      };
+      // No packages found in Publisher object - check if org has a registered namespace
+      const orgNamespace = orgDetails[0]?.NamespacePrefix;
+      if (orgNamespace && this.namespaces.has(orgNamespace)) {
+        // Org has a valid Omnistudio namespace registered
```
twu-sf, line 367: @sf-IrfanAmeeer-Shaik Outside of workspaces, orgs will not have omnistudio namespaces. Omnistudio namespaces only exists with the packages.

sf-IrfanAmeeer-Shaik: Yeah that's the intent  this only kicks in for internal workspace orgs like devops001gs0 that own the namespace directly instead of having a package installed. Customer orgs always go through the installed-package path, so this fallback never triggers for them.

twu-sf (approved):

PR #512, opened 2026-07-13
@W-23350384 - oclif node 24 25 compatibility

shaurabh-tiwari-git:
… assess

Fixes W-23350384: `sfdx omnistudio:migration:assess` crashes with `TypeError: this.config.plugins.filter is not a function` on Node 24/25 or sf CLI >= 2.14, and Node 25 hits a separate
`webidl.util.markAsUncloneable` failure in sf's bundled undici. Only the old combo (Node 22 + sf 2.12.7 + rc.61) worked pre-fix.

Root cause:
- `bin/run` bootstrapped via `@oclif/command@1` -> `@oclif/config@1`, which reads `config.plugins` as an Array. sf CLI >= 2.14 ships `@oclif/core >= 3.0` where `plugins` is a Map, so `.filter()` throws.
- The pinned `@oclif/core@4.8.0` resolution is also incompatible with Node 24.18+, surfacing as `Cannot read properties of undefined (reading 'ux')` in `sf-plugins-core/sfCommand.js`.

Changes:
- Rewrite `bin/run` to boot via `@oclif/core`'s `execute()` instead of the legacy `@oclif/command@1` runner.
- Bump `@oclif/core` resolution to `4.11.14` for Node 24 support.
- Bump `@oclif/plugin-command-snapshot` to `^5.3.30` and `@salesforce/plugin-command-reference` to `^3.1.119`.
- Drop the stale `minimatch` resolution that broke oclif-core's plugin-loader.
- Regenerate `command-snapshot.json` for the new flag discovery.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@oclif/config": "^1",
     "@oclif/errors": "^1",
     "@salesforce/core": "^8.23.4",
-    "@salesforce/sf-plugins-core": "^11.3.10",
+    "@salesforce/sf-plugins-core": "^11.3.12",
     "@types/jsdom": "^21.1.7",
     "@types/lodash.chunk": "^4.2.9",
     "@types/shelljs": "^0.8.15",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -75,7 +75,6 @@
     "js-yaml": "^4.1.1",
     "form-data": "^4.0.4",
     "trim-newlines": "^3.0.1",
-    "minimatch": "^3.1.4",
     "tough-cookie": "^4.1.3",
```
sf-kishore-kurri, line 78: @shaurabh-tiwari-git Can you update the README supported versions for sf cli, node and other dependencies to be installed on local workstation to avoid any errors

sf-pallavi-das (approved):
LGTM, please check if the readme needs update

package.json
shaurabh-tiwari-git: Addressed

snehaljha-sf (approved):

shaurabh-tiwari-git:
TODO: Validate the Windows version and update the README file.

